PR #8, opened 2019-03-06
verify-subtree.sh: relax check and ignore old content

pohly:
If for whatever reasons a repo already had a `release-tools` directory
before doing a clean import of it with `git subtree`, the check used
to fail because it found those old commits.

This can be fixed by telling `git log` to stop when the directory
disappears from the repo. There has to be a commit with removes the
old content, because otherwise `git subtree add` doesn't work.

Fixes: https://github.com/kubernetes-csi/external-resizer/issues/21

### DIFF
--- a/verify-subtree.sh
+++ b/verify-subtree.sh
@@ -30,7 +30,7 @@ if [ ! "$DIR" ]; then
     exit 1
 fi
 
-REV=$(git log -n1 --format=format:%H --no-merges -- "$DIR")
+REV=$(git log -n1 --remove-empty --format=format:%H --no-merges -- "$DIR")
 if [ "$REV" ]; then
     echo "Directory '$DIR' contains non-upstream changes:"
     echo


gnufied:
@pohly can you address comments here:

I have been thinking some more about it. Does it make sense to use a content based diff for verifying subtree changes rather than git log based one? Something like:

git remote add csi-release-tools_root https://github.com/kubernetes-csi/csi-release-tools.git
git fetch csi-release-tools_root
git diff remotes/csi-release-tools_root/master HEAD:release-tools/

It does mean all CSI repos must have latest version of release-tools.

pohly:
> It does mean all CSI repos must have latest version of release-tools.

Exactly, and that is the problem. Each time we update the csi-release-tools master, the PR checks in all repos using it will start failing. Same for all old branches which have to be on an older revision (for example, because master adds new source code checks that older release do not pass).

One could argue that this is intentional because it forces maintainers to drop everything they are doing and sync up with csi-release-tools. But I find that rather intrusive and for older branches it may not be right.

Instead, the git diff could be against the revision that was imported. But determining that can only be done my parsing the commit messages, and that is probably not easier than the current approach.

PR #10, opened 2019-03-27
check vendor directory

pohly:
This runs "dep check" to verify that the vendor directory is
up-to-date and meets expectations (= done with dep >= 0.5.0).


pohly:
Worked as expected in the other two repos.

PR #11, opened 2019-03-28
verify shellcheck

pohly:
This will be used for the Prow test script and could also be applied to other scripts, like for example the deploy.sh script in csi-driver-host-path.


util.sh
```diff
@@ -0,0 +1,839 @@
...
+  echo -n "${bin}"
+}
+
+kube::util::find-binary() {
+  kube::util::find-binary-for-platform "$1" "$(kube::util::host_platform)"
+}
+
+# Run all known doc generators (today gendocs and genman for kubectl)
+# $1 is the directory to put those generated documents
+kube::util::gen-docs() {
```
msau42: Many of these functions do not seem relevant for us. Can we remove the unused ones?

verify-shellcheck.sh
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env bash
```
msau42: Can you add a README describing what are we validating with this script?

util.sh
pohly: Okay.

verify-shellcheck.sh
pohly: Okay.

util.sh
pohly: I kept some smaller functions which may become useful.

pohly:
@msau42 updated.

PR #13, opened 2019-04-08
prow enhancements

pohly:
This is related to https://github.com/kubernetes/test-infra/pull/12088 but both can be merged independently.

/assign @msau42

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -203,7 +203,16 @@ configvar CSI_PROW_SANITY_CONTAINER "hostpath" "Kubernetes container with CSI dr
 # by setting CSI_PROW_TESTS_SANITY.
 configvar CSI_PROW_TESTS "unit parallel serial parallel-alpha serial-alpha ${CSI_PROW_TESTS_SANITY}" "tests to run"
 test_enabled () {
-    echo "${CSI_PROW_TESTS}" | grep -q -w -e "$1"
+    local test="$1"
+    # We want word-splitting here, so ignore: Double quote to prevent globbing and word splitting.
+    # shellcheck disable=SC2086
+    set ${CSI_PROW_TESTS}
+    for t in "$@"; do
+        if [ "$t" = "$test" ]; then
+            return
+        fi
+    done
+    return 1
 }
 
 # Serial vs. parallel is always determined by these regular expressions.
--- a/prow.sh
+++ b/prow.sh
@@ -619,18 +619,18 @@ install_sanity () (
 
 # Whether the hostpath driver supports raw block devices depends on which version
 # we are testing. It would be much nicer if we could determine that by querying the
-# installed driver.
+# installed driver's capabilities instead of having to do a version check.
 hostpath_supports_block () {
-    if [ -e "cmd/hostpathplugin" ] && ${CSI_PROW_BUILD_JOB}; then
-        # The assumption is that if we build the hostpath driver, then it is
-        # a current version with support.
-        echo true
-        return
-    fi
-
-    case "${CSI_PROW_DEPLOYMENT}" in kubernetes-1.13) echo false;; # wasn't supported and probably won't be backported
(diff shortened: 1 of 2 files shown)


pohly:
/hold

The version check is not quite right yet.

pohly:
KinD is failing in Prow for the "on-master" jobs (and only for those). Works for me locally. I would prefer to merge this PR as it is right now and investigate that failure separately.

pohly:
/hold

I'm starting to think that `AllAlpha=true` is the culprit. If so, one commit in this PR needs to be removed. Let's hold, I'm looking into that right now.

pohly:
/hold cancel

I've removed the `AllAlpha=true` and instead added a warning to not use it. kubelet accepted it, but then failed to run correctly.

pohly:
All four PRs which include this change ran okay:
- https://github.com/kubernetes-csi/csi-lib-utils/pull/26
- https://github.com/kubernetes-csi/external-attacher/pull/138
- https://github.com/kubernetes-csi/external-snapshotter/pull/111
- https://github.com/kubernetes-csi/node-driver-registrar/pull/37

prow.sh
```diff
@@ -846,102 +855,99 @@ main () {
         run_with_go "${CSI_PROW_GO_VERSION_BUILD}" make container || die "'make container' failed"
     fi
 
-    install_kind || die "installing kind failed"
-    start_cluster || die "starting the cluster failed"
+    if test_enabled "sanity" || test_enabled "parallel" || test_enabled "serial" || test_enabled "serial-alpha" || test_enabled "parallel-alpha"; then
```
msau42: How about having a couple of variables that can serve as whitelists of jobs that:

* require non-alpha cluster
* require alpha cluster

prow.sh
```diff
@@ -257,7 +257,10 @@ configvar CSI_PROW_E2E_ALPHA "$(get_versioned_variable CSI_PROW_E2E_ALPHA "${csi
 configvar CSI_PROW_E2E_ALPHA_GATES_1_13 'VolumeSnapshotDataSource=true,BlockVolume=true,CSIBlockVolume=true' "alpha feature gates for Kubernetes 1.13"
 configvar CSI_PROW_E2E_ALPHA_GATES_1_14 'VolumeSnapshotDataSource=true' "alpha feature gates for Kubernetes 1.14"
 # TODO: add new CSI_PROW_ALPHA_GATES entry for future Kubernetes releases
-configvar CSI_PROW_E2E_ALPHA_GATES_LATEST 'AllAlpha=true' "alpha feature gates for latest Kubernetes"
+# Careful with AllAlpha=true: enabling all alpha features turned out
+# to be problematic (kubelet died), so let's better list alpha gates
+# explicitly.
+configvar CSI_PROW_E2E_ALPHA_GATES_LATEST 'VolumeSnapshotDataSource=true' "alpha feature gates for latest Kubernetes"
```
msau42: ExpandCSIVolumes is also an alpha feature for 1.14.

prow.sh
pohly: You mean functions like `job_needs_cluster`, `job_needs_alpha_cluster`, `job_needs_non_alpha_cluster` instead of spelling that out via `test_enabled`? Yes, that would be a bit cleaner, albeit not necessarily shorter overall. I'll change it.

prow.sh
pohly: Ack.

pohly:
Pushed revised commits:
- cda2fc5: now with ExpandCSIVolumes
- 546d550: capture logs after cluster failure (suggestion from Ben)
- aa45a1c: avoids long list of `test_enabled` conditions

pohly:
All tests in the individual PRs passed, with just the known issue that hostpath v1.0.1 fails some csi-sanity tests. This will be fixed by updating the 1.13  deployment with hostpath v1.1.0.

PR #15, opened 2019-04-11
build.make: Replace 'return' to 'exit' to fix shellcheck error

pengzhisun:
I've got following error message while executed the `make test` command:

> Using shellcheck 0.6.0 docker image.
> Congratulations! All shell files are passing lint.
> /bin/sh: line 0: return: can only `return' from a function or sourced script
> make: *** [test-shellcheck] Error 1

The root cause is the test-shellcheck section in the build.make file should use `exit` command to return status code, instead of using `return`.

### DIFF
--- a/build.make
+++ b/build.make
@@ -143,11 +143,11 @@ test-shellcheck:
 	@ ret=0; \
 	if ! command -v docker; then \
 		echo "skipped, no Docker"; \
-		return 0; \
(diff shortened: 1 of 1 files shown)


pohly:
Thanks for the PR, it's good to see that this is getting used and people care to report and fix issues.

PR #19, opened 2019-05-10
build.make: allow repos to use 'go mod' for vendoring

pohly:
How a repo does vendoring is detected based on the presence of
Gopkg.toml.

The vendor check with `dep` was all done locally, but the
corresponding check for `go mod` requires network access. The check
therefore gets skipped when running in the Prow CI in situations where
we are sure that it isn't needed (for example, in a periodic job).

This is required for https://github.com/kubernetes-csi/csi-test/pull/188

```release-note
allow vendoring with 'go mod'
```
/kind feature

### DIFF
--- a/build.make
+++ b/build.make
@@ -118,14 +118,39 @@ test-fmt:
 	fi
 
 # This test only runs when dep >= 0.5 is installed, which is the case for the CI setup.
+# When using 'go mod', we allow the test to be skipped in the Prow CI under some special
+# circumstances, because it depends on accessing all remote repos and thus
+# running it all the time would defeat the purpose of vendoring:
+# - not handling a PR or
+# - the fabricated merge commit leaves go.mod, go.sum and vendor dir unchanged
+# - release-tools also didn't change (changing rules or Go version might lead to
+#   a different result and thus must be tested)
 .PHONY: test-vendor
 test: test-vendor
 test-vendor:
 	@ echo; echo "### $@:"
-	@ case "$$(dep version 2>/dev/null | grep 'version *:')" in \
-		*v0.[56789]*) dep check && echo "vendor up-to-date" || false;; \
-		*) echo "skipping check, dep >= 0.5 required";; \
-	esac
+	@ if [ -f Gopkg.toml ]; then \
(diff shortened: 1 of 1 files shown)


build.make
```diff
@@ -118,14 +118,39 @@ test-fmt:
...
+	  else \
+		echo "Repo uses 'go mod' for vendoring."; \
+		if [ "$${JOB_NAME}" ] && \
+                   ( [ "$${JOB_TYPE}" != "presubmit" ] || \
+                     [ $$(git diff "${PULL_BASE_SHA}..HEAD" -- go.mod go.sum vendor release-tools | wc -l) -eq 0 ] ); then \
+			echo "Skipping vendor check because the Prow pre-submit job does not change vendoring."; \
+		elif ! GO111MODULE=on go mod vendor; then \
+			echo "ERROR: vendor check failed."; \
+			false; \
+		elif [ $$(git status --porcelain=2 -- vendor | wc -l) -gt 0 ]; then \
```
hoyho: may get an error when execute `git status --porcelain=2 -- vendor | wc -l`  with different git version?

```bash
 [root@vm43 csi-test]# git status --porcelain=2 -- vendor | wc -l
error: option `porcelain' takes no value
usage: git status [options] [--] <pathspec>...

    -v, --verbose         be verbose
    -s, --short           show status concisely
    -b, --branch          show branch information
    --porcelain           machine-readable output
    --long                show status in long format (default)
    -z, --null            terminate entries with NUL
    -u, --untracked-files[=<mode>]
                          show untracked files, optional modes: all, normal, no. (Default: all)
    --ignored             show ignored files
    --ignore-submodules[=<when>]
                          ignore changes to submodules, optional when: all, dirty, untracked. (Default: all)
    --column[=<style>]    list untracked files in columns

0
```

pohly: Which version of git are you using?

We can drop the `=2`, the original format should be good enough. All we care for is "empty" vs. "not empty" anyway.

pohly: Thanks for testing, BTW.

pohly: I've removed the `=2`.

PR #20, opened 2019-07-08
more flexible test driver config

pohly:
This was tested in and is needed for https://github.com/kubernetes-csi/csi-driver-host-path/pull/64.

/assign @msau42 

```release-notes
NONE
```


prow.sh
```diff
@@ -673,6 +674,29 @@ hostpath_supports_block () {
     echo "${result:-true}"
 }
 
+# The default implementation of this function generates a external
```
msau42: Can this file be generated from the hostpath driver repo instead of in the general test framework?

pohly: That would imply also moving everything else related to hostpath out of csi-release-tools. I didn't do that originally because I thought it might be useful to share that default code between repos and/or branches of the hostpath driver repo. I'm not sure whether that reasoning really holds.

But even if it doesn't, that would be a bigger change than I am willing to risk right now. Let's keep it in mind for a future improvement and for now let's keep it here, okay?

msau42: That's fine. It just seems a bit cumbersome to update how we test hostpath driver this way.

PR #21, opened 2019-07-11
Add 1.15 jobs

msau42:
Updates prow.sh to use 1.15 hostpath deployment. As a result, the kind version also has to be upgraded to 0.4.0, which requires 2 major changes to our prow job:

* We need to use the new `kind.sigs.k8s.io/v1alpha3 Cluster` API object, which uses `kubeadm.k8s.io/v1beta2` which only works for 1.15+. As a result, our alpha jobs can only run on 1.15+
* kind 0.4.0 supports very specific patch images of 1.15, 1.14, 1.13, etc, which are not the ones we have configured in our prow jobs. As a result, I have added logic in the prow job to override the patch versions.

Also updates the config so that sidecar testing uses the version of the hostpath repo at v1.2.0-rc2.

```release-notes
Update kind to 0.4.0 to be able to run 1.15 jobs.
Fixes sidecar testing to use stable hostpath images.
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -183,7 +183,7 @@ configvar CSI_PROW_WORK "$(mkdir -p "$GOPATH/pkg" && mktemp -d "$GOPATH/pkg/csip
 #
 # When no deploy script is found (nothing in `deploy` directory,
 # CSI_PROW_HOSTPATH_REPO=none), nothing gets deployed.
-configvar CSI_PROW_HOSTPATH_VERSION fc52d13ba07922c80555a24616a5b16480350c3f "hostpath driver" # pre-1.1.0
+configvar CSI_PROW_HOSTPATH_VERSION "v1.2.0-rc2" "hostpath driver"
 configvar CSI_PROW_HOSTPATH_REPO https://github.com/kubernetes-csi/csi-driver-host-path "hostpath repo"
 configvar CSI_PROW_DEPLOYMENT "" "deployment"
 configvar CSI_PROW_HOSTPATH_DRIVER_NAME "csi-hostpath" "the driver (aka provisioner) name of the chosen hostpath driver"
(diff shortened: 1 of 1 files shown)


msau42:
Testing in hostpath repo: https://github.com/kubernetes-csi/csi-driver-host-path/pull/66
Testing in external provisioner repo: https://github.com/kubernetes-csi/external-provisioner/pull/313

msau42:
/hold
Tests are failing in external-provisioner WIP PR

msau42:
/hold cancel
tests are passing

prow.sh
```diff
@@ -151,10 +183,10 @@ configvar CSI_PROW_WORK "$(mkdir -p "$GOPATH/pkg" && mktemp -d "$GOPATH/pkg/csip
 #
 # When no deploy script is found (nothing in `deploy` directory,
 # CSI_PROW_HOSTPATH_REPO=none), nothing gets deployed.
-configvar CSI_PROW_HOSTPATH_VERSION fc52d13ba07922c80555a24616a5b16480350c3f "hostpath driver" # pre-1.1.0
+configvar CSI_PROW_HOSTPATH_VERSION "v1.2.0-rc2" "hostpath driver"
 configvar CSI_PROW_HOSTPATH_REPO https://github.com/kubernetes-csi/csi-driver-host-path "hostpath repo"
 configvar CSI_PROW_DEPLOYMENT "" "deployment"
-configvar CSI_PROW_HOSTPATH_DRIVER_NAME "csi-hostpath" "the driver (aka provisioner) name of the chosen hostpath driver"
+configvar CSI_PROW_HOSTPATH_DRIVER_NAME "hostpath.csi.k8s.io" "the hostpath driver name"
```
pohly: This works because v1.2.0-rc2 now uses that driver name for all deployments.  :+1:

PR #25, opened 2019-08-22
Update kind to v0.5.0

msau42:
We previously had to build kind off of a master commit to get a fix for K8s master node images.  Kind 0.5.0 has been released with the fix, so we can switch to the official release version now.

Tested in https://github.com/kubernetes-csi/csi-driver-host-path/pull/89


msau42:
Note that I'm not including the fix to ignore the new test case that requires SSH.  That is being addressed separately via https://github.com/kubernetes/kubernetes/pull/81755

PR #30, opened 2019-09-07
build windows binaries with .exe suffix

msau42:
so that it doesn't override linux binary


msau42:
Tested in https://github.com/kubernetes-csi/csi-driver-host-path/pull/93

ddebroy:
Given we don't have any requirements for non EXE binaries (like DLLs) this looks good.

PR #34, opened 2019-10-04
update Go mod support

pohly:
It turned out that changes like
https://github.com/kubernetes-csi/csi-lib-utils/pull/33 should better
have been committed after `go mod tidy` because that adds some
indirect dependencies in that example.

The revised `test-vendor` checks for that and (just in case that this
ever becomes desired) allows projects to not have a vendor directory
when using `go mod`.

How to use `go mod` properly gets documented in the README.md, because
there are such pitfalls.

### DIFF
--- a/go-get-kubernetes.sh
+++ b/go-get-kubernetes.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script can be used while converting a repo from "dep" to "go mod"
+# by calling it after "go mod init" or to update the Kubernetes packages
+# in a repo that has already been converted. Only packages that are
+# part of kubernetes/kubernetes and thus part of a Kubernetes release
+# are modified. Other k8.io packages (like k8s.io/klog, k8s.io/utils)
+# need to be updated separately.
+
+set -o pipefail
+
+cmd=$0
+
+function help () {
+    echo "$cmd <kubernetes version = x.y.z> - update all components from kubernetes/kubernetes to that version"
+}
+
+if [ $# -ne 1 ]; then
+    help
+    exit 1
+fi
+case "$1" in -h|--help|help) help; exit 0;; esac
+
+die () {
+    echo >&2 "$@"
+    exit 1
+}
+
+k8s="$1"
+
+# If the repo imports k8s.io/kubernetes (directly or indirectly), then
+# "go mod" will try to find "v0.0.0" versions because
+# k8s.io/kubernetes has those in it's go.mod file
+# (https://github.com/kubernetes/kubernetes/blob/2bd9643cee5b3b3a5ecbd3af49d09018f0773c77/go.mod#L146-L157).
+# (https://github.com/kubernetes/kubernetes/issues/79384).
+#
+# We need to replicate the replace statements to override those fake
+# versions also in our go.mod file (idea and some code from
+# https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597).
+mods=$( (set -x; curl --silent --show-error --fail "https://raw.githubusercontent.com/kubernetes/kubernetes/v${k8s}/go.mod") |
+          sed -n 's|.*k8s.io/\(.*\) => ./staging/src/k8s.io/.*|k8s.io/\1|p'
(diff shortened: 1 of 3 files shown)


pohly:
This shows that the revised check finds real problems (using csi-lib-utils as example): https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-lib-utils/34/pull-kubernetes-csi-csi-lib-utils/1180096611968094208

And this shows that the check passes after fixing the issue in csi-lib-utils: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-lib-utils/34/pull-kubernetes-csi-csi-lib-utils/1180097753288544256

pohly:
Hmm, as I try "go mod" in more complex repos (specifically, external-attacher) I'm finding that tracking dependencies on Kubernetes components without semantic version tags is a pain. With `dep`, we could list the desired `kubernetes-1.x.y` tags and do a search/replace followed by `dep ensure`. Not so with `go mod`. It doesn't even list the tag, so it's much harder to review whether this set of deps is consistent:
```
        k8s.io/api v0.0.0-20190614205929-e4e27c96b39a
        k8s.io/apimachinery v0.0.0-20190612125636-6a5db36e93ad
        k8s.io/client-go v11.0.0+incompatible
        k8s.io/csi-translation-lib v0.0.0-20190615091142-9ff632302e7e
        k8s.io/klog v0.3.2
        k8s.io/kube-openapi v0.0.0-20190401085232-94e1e7b7574c // indirect
        k8s.io/utils v0.0.0-20190308190857-21c4ce38f2a7 // indirect
```

`external-attacher` might be partly to blame for the inconsistency: it's `Gopkg.toml` doesn't have the necessary constraints for all components.

But even after a manual `GO111MODULE=on go get k8s.io/client-go@kubernetes-1.15.0 k8s.io/api@kubernetes-1.15.0 k8s.io/apimachinery@kubernetes-1.15.0` the result is still not nice:
```
        k8s.io/api v0.0.0-20190620084959-7cf5895f2711
        k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
        k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
        k8s.io/csi-translation-lib v0.0.0-20190615091142-9ff632302e7e
        k8s.io/klog v0.3.2
        k8s.io/kube-openapi v0.0.0-20190401085232-94e1e7b7574c // indirect
        k8s.io/utils v0.0.0-20190308190857-21c4ce38f2a7 // indirect
```

There's no indication of what the desired Kubernetes version is.

Is it worth writing a script for "update all Kubernetes components to a certain Kubernetes release"? That would simplify the update step, but result would be the same.

msau42:
Is there a way to specify/override constraints with tags? Maybe check with k8s-code-organization if they have any suggestions. Otherwise a script may be sufficient.

pohly:
One can and (as I noticed now) has to use `replace` statements. But those cannot use tag names. So one still needs a script to create those.

Looks like others have hit this before. For example, here's a script: https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597 ("after much weeping and gnashing of teeth", indeed!).

The one I came up with initially just used "go get" and thus hit that "unknown revision v0.0.0" error in repos which import `k8s.io/kubernetes` (e.g. `external-provisioner`). I incorporated the solution from that issue and then was able to convert and update both `external-attacher` and `external-provisioner` - see the new `go-get-kubernetes.sh`.

pohly:
> One can and (as I noticed now) has to use replace statements. But those cannot use tag names.

Or rather, one can use tag names there, but then `go mod tidy` will replace them with `v0.0.0-<revision>`.

go-get-kubernetes.sh
```diff
@@ -0,0 +1,104 @@
...
+case "$1" in -h|--help|help) help; exit 0;; esac
+
+die () {
+    echo >&2 "$@"
+    exit 1
+}
+
+k8s="$1"
+
+# If the repo imports k8s.io/kubernetes (directly or indirectly), then
```
msau42: Do we actually have repos that do this? I checked csi-lib-utils and external-provisioner, and didn't find any kubernetes/kubernetes dependencies.

pohly: `k8s.io/kubernetes` does get pulled in at the moment:
- https://github.com/kubernetes-csi/external-provisioner/tree/master/vendor/k8s.io/kubernetes
- https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/vendor/k8s.io/kubernetes

For the driver it is the mount code which still hasn't been moved into a staging repo (there is work under way, I just don't remember how far it got). Not sure for `external-provisioner`, haven't checked.

msau42: mount dependency is expected and will soon be removed. In general, we should start trying to cleanup and avoid any k8s.io/kubernetes dependencies.  But I guess it's fine to have this script in the meantime.

Does this help with the non-k8s.io/kubernetes repos (like k8s.io/api)?

pohly: k8s.io/api is from k/k and gets updated by the script.

k8s.io/util isn't, but doesn't tag releases. k8s.io/klog has separate releases and shows up in `go.mod` with a proper version number (`k8s.io/klog v0.3.1`). I think "go get" picks the version specified by some other go.mod file for those modules, just as it does for other modules outside of Kubernetes.

PR #36, opened 2019-10-08
Do full git clones in travis

msau42:


davidz627:
Do we have the root cause? "depth exactly 50" is suspicious or does this mean "depth >= 50 but we only ever get to 50 before failing". Besides that LGTM
@krzyzacy did you ever run into something like this before with PROW?

msau42:
cc @pohly 
I don't fully understand it either. Because https://github.com/kubernetes-csi/external-provisioner/pull/354 which went in right before mine succeeded.

pohly:
> Do  we have the root cause? "depth exactly 50" is suspicious or does this  mean "depth >= 50 but we only ever get to 50 before failing".

When "git clone --depth=50" happens to truncate exactly at a merge commit inside `release-tools`, then that commit is replaced with a grafted commit that has zero parents instead of the real two parents. That then is picked up by the test as an invalid modification of `release-tools`.

If we truncate at some other commit, the test passes.

> @krzyzacy did you ever run into something like this before with PROW?

Prow doesn't use shallow checkouts, so no.

Issue #24, opened 2019-08-13
prow.sh is missing kind cluster cleanup


msau42:
This is the current suspected reason why jobs are timing out

/assign

msau42:
Here's an example test run where the prow job timed out on teardown. https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-csi-canary-on-kubernetes-1-14/1195515914515124230

The recommendation I got from kind folks was to explicitly teardown the cluster instead of depending on prow to clean up the containers

PR #46, opened 2019-11-16
delete kind cluster after tests run

mucahitkurt:
Delete kind cluster after tests run. Inside a real Prow job it is better to clean up at runtime instead of leaving that to the Prow job cleanup code because the later sometimes times out.

Tested with the PR: https://github.com/kubernetes-csi/csi-driver-host-path/pull/120

/release-note-none

Fixes: https://github.com/kubernetes-csi/csi-release-tools/issues/24

cc @msau42 @pohly

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -580,6 +580,19 @@ EOF
     export KUBECONFIG
 }
 
+# Deletes kind cluster inside a prow job
+delete_cluster_inside_prow_job() {
+    # Inside a real Prow job it is better to clean up at runtime
+    # instead of leaving that to the Prow job cleanup code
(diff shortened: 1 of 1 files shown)


mucahitkurt:
> Can you change the commit message so that it follows the normal conventions (summary line, blank line, more detailed description)? That's better for humans (git log) and tools that work with commits, like the script that I intend to run.
> 
> That script summaries changes with the summary line, for example like this:
> [kubernetes-csi/csi-driver-host-path#115](https://github.com/kubernetes-csi/csi-driver-host-path/pull/115)
> 
> Everything in one long line will look odd.

@pohly I didn't know that, thank you! Fixed!

pohly:
/assign @msau42 

For final approval. If she doesn't comment this week, I'll do it.

msau42:
/approve
Thank you!

PR #47, opened 2019-11-19
Adding build for ppc64le

Pensu:
This PR adds builds for ppc64le.

```release-note
Cross build binaries for ppc64le
```

A relevant issue for discussion has been put here: https://github.com/kubernetes-csi/csi-release-tools/issues/48


build.make
```diff
@@ -70,6 +70,7 @@ build-%: check-go-version-go
 	CGO_ENABLED=0 GOOS=linux go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$* ./cmd/$*
 	if [ "$$ARCH" = "amd64" ]; then \
 		CGO_ENABLED=0 GOOS=windows go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*.exe ./cmd/$* ; \
+		CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*-ppc64le ./cmd/$*
```
Madhu-1: question: are we copying built all binaries to the same image or planning to build different docker images for different architecture maybe make use of `docker manifest ` feature?

Pensu: That would be the next step. This change only adds support for cross building binary, in future we would expect to build and push multi-arch images as well.

PR #53, opened 2019-12-04
fix syntax for ppc64le build

msau42:


ggriffiths:
Just tried this locally and it fixed the problem.
/lgtm

PR #57, opened 2019-12-21
Fix version_gt to work with Kubernetes prefix

ggriffiths:
Keeping this function in sync with the hostpath version.
https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/util/deploy-hostpath.sh#L104

Signed-off-by: Grant Griffiths <grant@portworx.com>


ggriffiths:
Testing changes in https://github.com/kubernetes-csi/csi-driver-host-path/pull/131

PR #61, opened 2020-01-02
Fix overriding of junit results

msau42:


msau42:
Tested in https://github.com/kubernetes-csi/csi-driver-host-path/pull/133 and
https://github.com/kubernetes-csi/external-resizer/pull/68/files

PR #65, opened 2020-01-10
Update hostpath driver version to get fix for connection-timeout

msau42:
Testing in https://github.com/kubernetes-csi/livenessprobe/pull/59


ggriffiths:
checked that new tag exists:
https://github.com/kubernetes-csi/csi-driver-host-path/releases/tag/v1.3.0-rc3

and looks like kubernetes-csi/livenessprobe#59 is passing with this change.
/lgtm

PR #69, opened 2020-02-10
prow.sh: generic driver installation

pohly:
This relies on a slightly different deployment script: a "deploy.sh"
must exist which knows that it has to dump a test driver configurion
into the file pointed to with CSI_PROW_TEST_DRIVER, if that env
variable is set.

That way, we no longer need to know what capabilities the installed
driver has.

The corresponding change in the hostpath repo is in https://github.com/kubernetes-csi/csi-driver-host-path/pull/143


msau42:
/lgtm
Thanks this is great!

PR #73, opened 2020-02-20
Adding s390x cross build support

srajmane:
Adding cross build binary support for s390x.

### DIFF
--- a/build.make
+++ b/build.make
@@ -71,6 +71,7 @@ build-%: check-go-version-go
 	if [ "$$ARCH" = "amd64" ]; then \
 		CGO_ENABLED=0 GOOS=windows go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*.exe ./cmd/$* ; \
 		CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*-ppc64le ./cmd/$* ; \
+		CGO_ENABLED=0 GOOS=linux GOARCH=s390x go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*-s390x ./cmd/$* ; \
 	fi
 
 container-%: build-%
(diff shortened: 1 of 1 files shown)


srajmane:
@pohly @msau42 Could you please have a look ?

srajmane:
@pohly @msau42 Could you please have a look ?
Corrected syntax  for ARCH detection, observed that current if block is not creating binaries for mentioned architecture.
Updated **if condition** and verified it manually .

pohly:
The change itself looks good, I'm just wondering whether should continue to add special build rules like this for different target architectures. Should we just add them for now when people ask for it, or should we stop and try to find a more generic approach, in particular one which then also covers building and publishing of images?

As I don't have the answers for those problems, I myself lean towards simply merging this.

namrata-ibm:
@pohly Thanks for your comments. Yes we are also planning to work on generic images, however for now, adding above would be helpful.

build.make
```diff
@@ -68,9 +68,10 @@ ARCH := $(if $(GOARCH),$(GOARCH),$(shell go env GOARCH))
 build-%: check-go-version-go
 	mkdir -p bin
 	CGO_ENABLED=0 GOOS=linux go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$* ./cmd/$*
-	if [ "$$ARCH" = "amd64" ]; then \
+	if [ "$(ARCH)" = "amd64" ]; then \
 		CGO_ENABLED=0 GOOS=windows go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*.exe ./cmd/$* ; \
 		CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*-ppc64le ./cmd/$* ; \
+		CGO_ENABLED=0 GOOS=linux GOARCH=s390x go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*-s390x ./cmd/$* ; \
```
pohly: Instead of forcing all developers to build for all of these architectures, let's try something else:
- by default, build without overriding GOOS and GOARCH
- add some way (env vars?!) of enabling the building of some other versions of the binaries
- enable that building in the CI (= prow.sh)

Let me try to come up with such a solution.

pohly: See https://github.com/kubernetes-csi/csi-release-tools/pull/76

srajmane:
Closing in favor of #76

PR #74, opened 2020-02-21
Adding s390x travis support

srajmane:
- Added s390x support for travis
- Disable shellcheck for s390x
- Added docker image name with arch name.


srajmane:
@pohly @msau42 Could you please have a look ?

msau42:
/assign @pohly 
I believe we want to move away from travis and start migrating our release infra to the system supported by wg-k8s-infra.

pohly:
Let's focus on the approach outlined in https://github.com/kubernetes-csi/node-driver-registrar/pull/63 instead. That also is open at this time, but it looks more promising than extending the build rules for every single additional architecture and continuing to rely on TravisCI.

srajmane:
Closing in favor of Prow builds.

PR #75, opened 2020-03-03
kind v0.7.0, revised CSI_PROW_KUBERNETES_VERSION handling

pohly:
The first commit fixes a usability issue when invoking the script manually. The second updates to kind 0.7.0 (the current version), and the last one acknowledges that the specific selection of the Kubernetes version (including the patch version number) must be done inside the component's prow script and not in the test-infra job spec.

That was already introduced for 1.15.x earlier out of necessity, now it is done for all releases and the semantic of CSI_PROW_KUBERNETES_VERSION gets changed accordingly.

Once this is rolled out in all repos and all relevant branches, the test-infra jobs can be simplified to just use `CSI_PROW_KUBERNETES_VERSION=x.y`.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -525,6 +525,9 @@ start_cluster () {
         run kind delete cluster --name=csi-prow || die "kind delete failed"
     fi
 
+    # With kind >= 0.6.0 we can choose where to put the resulting configuration.
+    export KUBECONFIG="${CSI_PROW_WORK}/kube.config"
+
     # Build from source?
     if [[ "${CSI_PROW_KUBERNETES_VERSION}" =~ ^release-|^latest$ ]]; then
         if ! ${csi_prow_kind_have_kubernetes}; then
@@ -608,7 +611,6 @@ EOF
             die "Cluster creation failed again, giving up. See the 'kind-cluster' artifact directory for additional logs."
         fi
     fi
-    export KUBECONFIG="${HOME}/.kube/config"
 }
 
 # Deletes kind cluster inside a prow job
--- a/prow.sh
+++ b/prow.sh
@@ -42,6 +42,19 @@
 RELEASE_TOOLS_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")")"
 REPO_DIR="$(pwd)"
 
+info () {
+    echo >&2 INFO: "$@"
+}
+
+warn () {
+    echo >&2 WARNING: "$@"
+}
+
+die () {
+    echo >&2 ERROR: "$@"
+    exit 1
+}
+
 # Sets the default value for a variable if not set already and logs the value.
 # Any variable set this way is usually something that a repo's .prow.sh
 # or the job can set.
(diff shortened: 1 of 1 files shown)


pohly:
Test builds are here:
- https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/418/pull-kubernetes-csi-external-provisioner-1-15-on-kubernetes-1-15/1234571965780463616/build-log.txt
- https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/418/pull-kubernetes-csi-external-provisioner-1-16-on-kubernetes-1-16/1234571965780463617/build-log.txt
- https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/418/pull-kubernetes-csi-external-provisioner-1-17-on-kubernetes-1-17/1234571965780463618/build-log.txt

prow.sh
```diff
@@ -107,7 +135,24 @@ configvar CSI_PROW_GO_VERSION_GINKGO "${CSI_PROW_GO_VERSION_BUILD}" "Go version
...
+# Kubernetes release. Those image tags may get replaced in the future
+# with something that is incompatible with kind release that we are
+# using, therefore reference images with the digest. We need at least
+# one image digest for each major.minor Kubernetes version
+# number. Listing more than one patch version is okay but not
+# necessary, because always the most recent one will be used.
+#
+# Digests are part of the release notes at: https://github.com/kubernetes-sigs/kind/releases
+configvar CSI_PROW_KIND_VERSION "v0.7.0" "kind"
+configvar CSI_PROW_KIND_IMAGE_DIGEST_1_17_0 "sha256:9512edae126da271b66b990b6fff768fbb7cd786c7d39e86bdf55906352fdf62" "kind image digest for Kubernetes 1.17.0"
```
msau42: Can we dynamically query this information? I want to start moving towards a goal where we  no longer need to update the testing framework for every new k8s release.

prow.sh
```diff
@@ -540,7 +568,18 @@ start_cluster () {
...
     else
-        image="kindest/node:v${CSI_PROW_KUBERNETES_VERSION}"
+        local varname
+        varname="CSI_PROW_KIND_IMAGE_DIGEST_$(echo "${CSI_PROW_KUBERNETES_VERSION}" | tr . _)"
+        local digest
+        digest=$(eval echo "\$$varname")
+        if [ "$digest" ]; then
+            image="kindest/node:v${CSI_PROW_KUBERNETES_VERSION}@$digest"
+        else
+            # We shouldn't get here because of override_k8s_version, but let's keep the
```
msau42: do we not have a canary job where we run against k8s master?

prow.sh
pohly: > Can we dynamically query this information?

It comes from https://github.com/kubernetes-sigs/kind/releases, which isn't that easily parsed. The format might also change between releases.

prow.sh
pohly: Yes. That's covered by the preceding ` if [[ "${CSI_PROW_KUBERNETES_VERSION}" =~ ^release-|^latest$ ]]; then` which picks the locally built `csiprow/node:latest` as image.

prow.sh
```diff
@@ -52,23 +65,38 @@ configvar () {
...
-# $1 if they are equal minor versions. Ignores versions that begin with
-# "release-".
+# Updates the patch version of CSI_PROW_KUBERNETES_VERSION such that we
+# have a kind image for it. Does nothing if the current value isn't a
+# version number.
 override_k8s_version () {
-    local current_minor_version
-    local override_minor_version
+    # Not a version number?
+    if ! echo "$CSI_PROW_KUBERNETES_VERSION" | grep -q '^[0-9][0-9]*\.[0-9][0-9]*'; then
```
msau42: valid values also include "release-1.17", etc

prow.sh
msau42: Instead of embedding config into our infrastructure and needing to update and sync it across all repos every release, can we do something instead like require the caller to pass in the kind version + k8s version (including patch)?

pohly: That then makes it harder to run tests locally because the repo is no longer self-contained. At least I test that way from time to time.

There's also the potential problem that updating to, say, kind 0.8.0 will require changes that go beyond increasing the version number and the image digests. We had that with the config handling in the kind 0.5.0 -> 0.6.0 transition. We would have to move the entire `start_cluster` and `install_kind` functions into `test-infra`.

prow.sh
pohly: That's exactly why this check is here: it makes `override_k8s_version` return early because there is nothing to be patched in `release-1.17`. For those versions we'll build the image ourselves instead of using a pre-built one.

pohly:
@msau42: shall I close this PR, as it doesn't seem to go in the direction where you want to go with setting up the KinD cluster?

prow.sh
msau42: I hope over time kind API will stabilize and we won't have this problem as frequently.

For ease of local testing, what if we just have one default? And if another version wants to be used, then the person needs to go to the kind releases to figure out the correct combinations. Or it falls back to building a node image.

I think it would also be good to ask kind folks if there's a way to make prebuild images more easily discoverable

PR #76, opened 2020-03-04
build for multiple platforms only in CI, add s390x

pohly:
Developers should not be forced to build for all platforms by
default. We also don't want to copy-and-paste the go invocation for
each new platform.

To address both, the target platform(s) are now configurable via
BUILD_PLATFORMS and additional platforms are only enabled in the Prow
CI.

For now this serves as a test that the source actually compiles for
multiple platforms. Building images for different target platforms is a
different problem.

### DIFF
--- a/build.make
+++ b/build.make
@@ -60,18 +60,25 @@ else
 TESTARGS =
 endif
 
-ARCH := $(if $(GOARCH),$(GOARCH),$(shell go env GOARCH))
-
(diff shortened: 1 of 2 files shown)


pohly:
This replaces PR #73 .

/release-note-none

pohly:
Test build is here: https://prow.k8s.io/log?id=1235150069494190080&job=pull-kubernetes-csi-external-provisioner-unit

PR #78, opened 2020-04-16
Fix csi-snapshotter RBAC yaml version

ggriffiths:
Caught in this PR:
https://github.com/kubernetes-csi/external-snapshotter/pull/298

Testing this change in this PR:
https://github.com/kubernetes-csi/external-snapshotter/pull/299

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -1064,18 +1064,24 @@ main () {
                 # always pulling the image
                 # (https://github.com/kubernetes-sigs/kind/issues/328).
                 docker tag "$i:latest" "$i:csiprow" || die "tagging the locally built container image for $i failed"
-            done
 
-            if [ -e deploy/kubernetes/rbac.yaml ]; then
-                # This is one of those components which has its own RBAC rules (like external-provisioner).
-                # We are testing a locally built image and also want to test with the the current,
-                # potentially modified RBAC rules.
-                if [ "$(echo "$cmds" | wc -w)" != 1 ]; then
-                    die "ambiguous deploy/kubernetes/rbac.yaml: need exactly one command, got: $cmds"
+                # For components with multiple cmds, the RBAC file should be in the following format:
+                #   rbac-$cmd.yaml
+                # If this file cannot be found, we can default to the standard location:
+                #   deploy/kubernetes/rbac.yaml
+                rbac_file_path=$(find . -type f -name "rbac-$i.yaml")
+                if [ "$rbac_file_path" == "" ]; then
+                    rbac_file_path="$(pwd)/deploy/kubernetes/rbac.yaml"
                 fi
-                e=$(echo "$cmds" | tr '[:lower:]' '[:upper:]' | tr - _)
-                images="$images ${e}_RBAC=$(pwd)/deploy/kubernetes/rbac.yaml"
-            fi
+                
+                if [ -e "$rbac_file_path" ]; then
+                    # This is one of those components which has its own RBAC rules (like external-provisioner).
+                    # We are testing a locally built image and also want to test with the the current,
+                    # potentially modified RBAC rules.
+                    e=$(echo "$i" | tr '[:lower:]' '[:upper:]' | tr - _)
+                    images="$images ${e}_RBAC=$rbac_file_path"
+                fi
+            done
         fi
 
         if tests_need_non_alpha_cluster; then


prow.sh
```diff
@@ -1066,15 +1066,21 @@ main () {
                 docker tag "$i:latest" "$i:csiprow" || die "tagging the locally built container image for $i failed"
             done
 
-            if [ -e deploy/kubernetes/rbac.yaml ]; then
+            if [ -e "$(pwd)/deploy/kubernetes/rbac.yaml" ] || [ -e "$(pwd)/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml" ]; then
```
pohly, line 1069: Why the additional `$(pwd)`? That's redundant here.

prow.sh
```diff
@@ -1066,15 +1066,21 @@ main () {
...
                 # This is one of those components which has its own RBAC rules (like external-provisioner).
                 # We are testing a locally built image and also want to test with the the current,
                 # potentially modified RBAC rules.
+
+                # snapshot-controller RBAC is handled by the cluster manager
```
pohly, line 1074: Can you add a link here that provides more details for how the cluster manager does that?

prow.sh
```diff
@@ -1066,15 +1066,21 @@ main () {
...
+                cmds=${cmds//snapshot-controller/}
                 if [ "$(echo "$cmds" | wc -w)" != 1 ]; then
                     die "ambiguous deploy/kubernetes/rbac.yaml: need exactly one command, got: $cmds"
                 fi
                 e=$(echo "$cmds" | tr '[:lower:]' '[:upper:]' | tr - _)
                 images="$images ${e}_RBAC=$(pwd)/deploy/kubernetes/rbac.yaml"
+                if [ -e "$(pwd)/deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml" ]; then
```
pohly, line 1081: Same redundant `$(pwd)` here.

prow.sh
msau42: I think this doesn't quite work for snapshotter repo which has two binaries with two rbac files.

Maybe we need to recursively search for rbac files and apply all of them?

prow.sh
msau42: In the case of a pull job, we want to test all local changes from the pr, so I think we still want to apply local rbac changes

ggriffiths: Yeah, it's mentioned in the new readme. 
https://github.com/kubernetes-csi/external-snapshotter/blob/master/README.md#overview
> The snapshot controller is deployed by the Kubernetes distributions and is responsible for watching the VolumeSnapshot CRD objects and manges the creation and deletion lifecycle of snapshots.

Here's where prow.sh installs the snapshot-controller and it's RBAC:
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L727

This happens right after cluster setup, along with the necessary CRD.
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L1089
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L1146

prow.sh
ggriffiths: I thought the $(pwd) was needed for this test: `pull-kubernetes-csi-external-snapshotter-1-16-on-kubernetes-1-16`. This was failing to hit this conditional originally. Let me remove it temporarily and see if that's still the case still.

Snapshotter 1.x rbac file is always `deploy/kubernetes/rbac.yaml`. The reason for needing an additional check here is for 2.x compatibility, where the csi-snapshotter RBAC is here: `deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml`

Regarding the snapshot-controller RBAC, it is taken care of here separately:
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L1089
https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L727

ggriffiths: I was mistaken, looks like $(pwd) isn't needed. Will remove now.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-snapshotter/299/pull-kubernetes-csi-external-snapshotter-1-16-on-kubernetes-1-16/1250813186752385026

ggriffiths: ok, the problem was actually the quotations around `deploy/kubernetes/rbac.yaml`, not the $(pwd). I just removed the quotations and it did not enter the conditional.

prow.sh
ggriffiths: > In the case of a pull job, we want to test all local changes from the pr, so I think we still want to apply local rbac changes

Ah yeah, will change this to handle the snapshot-controller.

msau42: discussed offline, snapshot-controller will be addressed separately. so this pr should just fix csi-snapshotter sidecar testing

prow.sh
```diff
@@ -1066,15 +1066,21 @@ main () {
                 docker tag "$i:latest" "$i:csiprow" || die "tagging the locally built container image for $i failed"
             done
 
-            if [ -e deploy/kubernetes/rbac.yaml ]; then
+            if [ -e "deploy/kubernetes/rbac.yaml" ] || [ -e "deploy/kubernetes/csi-snapshotter/rbac-csi-snapshotter.yaml" ]; then
```
msau42, line 1069: I think we will have other repos in the future that contain multiple binaries with different rbac files. Do you see a way that we can make the path configurable so that it will work for other repos?

Maybe we loop over $cmds and try to find rbac files that match?

prow.sh
ggriffiths: Issue for this: https://github.com/kubernetes-csi/csi-release-tools/issues/79

prow.sh
pohly, line 1069: This is still odd. Why do you need quotation marks here? It shouldn't make any difference whatsoever.

```
/tmp/empty$ touch foobar
/tmp/empty$ if [ -e foobar ]; then echo foobar exists; fi
foobar exists
/tmp/empty$ if [ -e "foobar" ]; then echo foobar exists; fi
foobar exists
```

ggriffiths: my mistake, I was updating this in my release-tools repo and testing in the sub-tree snapshotter version. You're right, it does not make a difference. Will update this too.

ggriffiths: @msau42 Updated this to find an RBAC file in the format `rbac-${cmd}.yaml`

Is this a format we can assume? I could generalize it to allow for `${cmd}-rbac.yaml` and other cases, but I'm not sure how far we want to go in generalizing it.

prow.sh
```diff
@@ -1064,18 +1064,27 @@ main () {
...
-                e=$(echo "$cmds" | tr '[:lower:]' '[:upper:]' | tr - _)
-                images="$images ${e}_RBAC=$(pwd)/deploy/kubernetes/rbac.yaml"
-            fi
+                
+                if [ -e "$rbac_file_path" ]; then
+                    # This is one of those components which has its own RBAC rules (like external-provisioner).
+                    # We are testing a locally built image and also want to test with the the current,
+                    # potentially modified RBAC rules.
+
+                    # snapshot-controller RBAC is handled by the cluster manager
```
msau42, line 1082: does this comment still apply?

prow.sh
msau42: I think this format is fine. Thanks!

prow.sh
ggriffiths: It does not. Removed

PR #82, opened 2020-04-29
Skip prow job when on Kubernetes v1.16.

bertinatto:
Going forward we are using the GA version of the CSINode object in the sidecars. However, Kubernetes v1.16 only contains the v1beta1 version of that object, hence we need to skip the prow job in this version.

/assign @pohly @msau42

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -1022,6 +1022,28 @@ function version_gt() {
 }
 
 main () {
+    # This check assumes that the current configuration uses a driver deployment
+    # which has been updated to use v1 APIs that aren't available in Kubernetes < 1.17.
+    # The check can be removed when all Prow jobs for Kubernetes < 1.17 are removed.
+    if ! ( version_gt "${CSI_PROW_KUBERNETES_VERSION}" "1.16.255" || "${CSI_PROW_KUBERNETES_VERSION}" == "latest" ); then
+	filtered_tests=
+	skipped_tests=
+	for test in ${CSI_PROW_TESTS}; do
+	    case "$test" in
+            parallel|parallel|serial|parallel-alpha|serial-alpha):
+			skipped_tests="$skipped_tests $test"
+		        ;;
+		*)
+			filtered_tests="$filterered_tests $test"
+			;;
+	    esac
+	done
+	if [ "$skipped_tests" ]; then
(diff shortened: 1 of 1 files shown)


bertinatto:
@pohly can you take a look if this is the right way to do it?

Hiding the whitespace changes makes it more readable: https://github.com/kubernetes-csi/csi-release-tools/pull/82/files?diff=unified&w=1

pohly:
Which repos are hitting this problem? All? Then a generic solution in prow.sh is the way to go.

However, I'd do it differently and filter CSI_PROW_TESTS. Then we don't need this indention change.

Something like:
```
func main() {
   # This check assumes that the current configuration uses a driver deployment
   # which has been updated to use v1 APIs that aren't available in Kubernetes < 1.17.
   # The check can be removed when all Prow jobs for Kubernetes < 1.17 are removed.
   if ! ( version_gt "${CSI_PROW_KUBERNETES_VERSION}" "1.16.255" || ${CSI_PROW_KUBERNETES_VERSION}" == "latest" ); then
    filtered_tests=
    skipped_tests=
    for test in ${CSI_PROW_TESTS}; do
       case $test in
         parallel|parallel|serial|parallel-alpha|serial-alpha) skipped_tests="$skipped_tests $test";;
         *) filtered_tests="$filterered_tests $test";;
       esac
    done
    if [ "$skipped_tests" ]; then
        info "Testing on Kubernetes ${CSI_PROW_KUBERNETES_VERSION} is no longer supported. Skipping CSI_PROW_TESTS $skipped_tests."
        CSI_PROW_TESTS="$filtered_tests"
    fi
  fi
```

msau42:
It's technically only needed for provisioner and attacher. I think we should probably override it in the repo's '.prow.sh'

pohly:
> It's technically only needed for provisioner and attacher.

Only at the moment, right? Once those sidecars have been released, are used in csi-driver-host-path, and csi-release-tools gets updated to use that updated csi-driver-host-path, then it will be needed by all components.

But I agree, for now it seems better to just do this filtering in the .prow.sh of these two components.

bertinatto:
> Only at the moment, right? Once those sidecars have been released, are used in csi-driver-host-path, and csi-release-tools gets updated to use that updated csi-driver-host-path, then it will be needed by all components.
> 
> But I agree, for now it seems better to just do this filtering in the .prow.sh of these two components.

OK, closing this in favor of moving the code to provisioner and attacher respectives `.prow.sh` files.

/close

PR #84, opened 2020-05-12
Fix markdown errors on the *.md files

humblec:

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,22 +6,23 @@ in the top-level Makefile.
 
 The rules include support for building and pushing Docker images, with
 the following features:
- - one or more command and image per project
- - push canary and/or tagged release images
- - automatically derive the image tag(s) from repo tags
- - the source code revision is stored in a "revision" image label
- - never overwrites an existing release image
 
-Usage
------
+- one or more command and image per project
+- push canary and/or tagged release images
+- automatically derive the image tag(s) from repo tags
+- the source code revision is stored in a "revision" image label
+- never overwrites an existing release image
+
+## Usage
 
 The expected repository layout is:
- - `cmd/*/*.go` - source code for each command
- - `cmd/*/Dockerfile` - docker file for each command or
+
+- `cmd/*/*.go` - source code for each command
+- `cmd/*/Dockerfile` - docker file for each command or
    Dockerfile in the root when only building a single command
- - `Makefile` - includes `release-tools/build.make` and sets
+- `Makefile` - includes `release-tools/build.make` and sets
    configuration variables
- - `.travis.yml` - a symlink to `release-tools/.travis.yml`
+- `.travis.yml` - a symlink to `release-tools/.travis.yml`
 
 To create a release, tag a certain revision with a name that
 starts with `v`, for example `v1.0.0`, then `make push`
@@ -35,8 +36,7 @@ Release branches are expected to be named `release-x.y` for releases
 `x.y.z`. Building from such a branch creates `x.y-canary`
 images. Building from master creates the main `canary` image.
 
-Sharing and updating
---------------------
+## Sharing and updating
 
 [`git subtree`](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt)
 is the recommended way of maintaining a copy of the rules inside the
@@ -50,8 +50,7 @@ Cheat sheet:
 - `git subtree pull --prefix=release-tools https://github.com/kubernetes-csi/csi-release-tools.git master` - update local copy to latest upstream (whenever upstream changes)
 - edit, `git commit`, `git subtree push --prefix=release-tools git@github.com:<user>/csi-release-tools.git <my-new-or-existing-branch>` - push to a new branch before submitting a PR
 
-verify-shellcheck.sh
---------------------
+## verify-shellcheck.sh script
 
 The [verify-shellcheck.sh](./verify-shellcheck.sh) script in this repo
 is a stripped down copy of the [corresponding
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,7 +51,6 @@ Cheat sheet:
 - edit, `git commit`, `git subtree push --prefix=release-tools git@github.com:<user>/csi-release-tools.git <my-new-or-existing-branch>` - push to a new branch before submitting a PR
 
 verify-shellcheck.sh
```
pohly, line 53: This need to be a sub-section.

README.md
```diff
@@ -78,12 +76,13 @@ main
 
 All Kubernetes-CSI repos are expected to switch to Prow. For details
 on what is enabled in Prow, see
-https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-csi
+<https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-csi>
```
pohly, line 79: Here the linter seems a bit too strict, given that GitHub (our main target) can handle plain embedded URLs just fine.

README.md
```diff
@@ -78,12 +76,13 @@ main
...
 
 Test results for periodic jobs are visible in
-https://testgrid.k8s.io/sig-storage-csi-ci
+<https://testgrid.k8s.io/sig-storage-csi-ci>
```
pohly, line 82: Same here. This check is likely to create a lot of churn in various repos if enabled.

README.md
```diff
@@ -98,17 +97,18 @@ When it terminates, the following command can be used to get access to
 the Kubernetes cluster that was brought up for testing (assuming that
 this step succeeded):
 
-    export KUBECONFIG="$(kind get kubeconfig-path --name="csi-prow")"
+`export KUBECONFIG="$(kind get kubeconfig-path --name="csi-prow")"`
```
pohly, line 100: Three backticks with `console` as type is probably better here.

README.md
```diff
@@ -98,17 +97,18 @@ When it terminates, the following command can be used to get access to
...
 
 It is possible to control the execution via environment variables. See
 `prow.sh` for details. Particularly useful is testing against different
 Kubernetes releases:
 
-    CSI_PROW_KUBERNETES_VERSION=1.13.3 ./.prow.sh
-    CSI_PROW_KUBERNETES_VERSION=latest ./.prow.sh
+```shell
```
pohly, line 106: Do we get better rendering with "console" instead of "shell"? "console" does syntax highlighting; I don't know whether "shell" does.

README.md
```diff
@@ -143,12 +144,14 @@ longer deemed necessary, then a project can also remove the directory.
 
 Conversion of a repository that uses `dep` to `go mod` can be done with:
 
+```shell
```
pohly, line 147: Please remove the indention inside the block.

SIDECAR_RELEASE_PROCESS.md
```diff
@@ -48,24 +49,29 @@ naming convention `<hostpath-deployment-version>-on-<kubernetes-version>`.
...
 1. Identify all issues and ongoing PRs that should go into the release, and
   drive them to resolution.
 1. Download v2.8+ [K8s release notes
   generator](https://github.com/kubernetes/release/tree/master/cmd/release-notes)
 1. Generate release notes for the release. Replace arguments with the relevant
   information.
     * For new minor releases on master:
-        ```
+
+      ```shell
```
pohly, line 61: Probably should remove indention here, too.

SIDECAR_RELEASE_PROCESS.md
```diff
@@ -48,24 +49,29 @@ naming convention `<hostpath-deployment-version>-on-<kubernetes-version>`.
...
         GITHUB_TOKEN=<token> release-notes --discover=mergebase-to-latest
         --github-org=kubernetes-csi --github-repo=external-provisioner
         --required-author="" --output out.md
         ```
+
     * For new patch releases on a release branch:
-        ```
+
+        ```shell
```
pohly, line 69: And here.

README.md
humblec: Done.

README.md
humblec: True. but the extra brackets are only added in the README of this repo. so it should not cause extra churn in important repos unless we enable mdlinter. Even then we can define a skip rule.

README.md
humblec: same as above.

README.md
humblec: done.

README.md
humblec: done

SIDECAR_RELEASE_PROCESS.md
humblec: Done.

SIDECAR_RELEASE_PROCESS.md
humblec: Done.

README.md
humblec: Done.

humblec:
@pohly apologies for the delay in reverting. I have addressed the comments, ptal. Thanks in advance.

humblec:
@msau42 :)

pohly:
Are the updated .md files now still passing the proposed linter check?

I think I would prefer to merge the .md changes and the linter in the same PR *and* see a similar test PR against some other repo with more documentation. That way, we will get a feeling for the changes in the .md files that will be required when enabling the linter, and we can tweak the linter when it enforces things that we don't care about. For example, plain URLs IMHO are fine. There's no need to enforce angular brackets around them, that's just causing churn.

PR #87, opened 2020-05-28
cloud build: initial set of shared files

pohly:
The approach taken here extends the existing support for
cross-compiling binaries on the build host and specifying the Go
compiler: Go is installed if needed (as in Prow testing), binaries are
build on the host, then one image is created for each platform, and
finally those are combined into a single multi-architecture image.

```release-note
The Dockerfile must accept a "binary" build argument.
```

### DIFF
--- a/build.make
+++ b/build.make
@@ -105,6 +105,69 @@ build: $(CMDS:%=build-%)
 container: $(CMDS:%=container-%)
 push: $(CMDS:%=push-%)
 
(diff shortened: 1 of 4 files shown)


pohly:
This was originally tested in node-driver-registrar (https://github.com/kubernetes-csi/node-driver-registrar/pull/84) and now just copies the files to csi-release-tools for sharing among repos.

PR #89, opened 2020-06-04
cloud build: initialize support for running commands in Dockerfile

pohly:
If the Dockerfile needs to run some command, that step fails unless
QEMU is set up properly first:
      failed to solve: rpc error: code = Unknown desc = failed to load
      LLB: runtime execution on platform linux/ppc64le not supported

/release-note-none
/cc @jsafrane


pohly:
Tested in csi-driver-host-path, now needs to be made available through csi-release-tools.

PR #96, opened 2020-08-10
update go version to 1.14.4

humblec:

### DIFF
--- a/travis.yml
+++ b/travis.yml
@@ -6,7 +6,7 @@ git:
   depth: false
 matrix:
   include:
-  - go: 1.13.3
+  - go: 1.14.4
 before_script:
 - mkdir -p bin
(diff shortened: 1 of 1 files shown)


pohly:
I don't think we came to a full conclusion on which Go version we want to switch to:
https://kubernetes.slack.com/archives/C8EJ01Z46/p1596537153179300

Apparently 1.14 has potential runtime issues that are severe enough that Kubernetes is skipping it.

pohly:
We decided to go straight for 1.15, in order to be aligned with Kubernetes 1.19.0 and Docker which both skip 1.14 because of runtime issues (https://github.com/kubernetes/kubernetes/issues/92521).

pohly:
PR #97 replaces this one here.

pohly:
/close

We went straight to Go 1.15 instead.

PR #97, opened 2020-08-12
switch to Go 1.15

pohly:
Go 1.15 was released and is the major version that Kubernetes 1.19.0
is going to use. There are probably bugs in the older 1.13.3 that were
fixed, so we should update.

### DIFF
--- a/travis.yml
+++ b/travis.yml
@@ -6,7 +6,7 @@ git:
   depth: false
 matrix:
   include:
-  - go: 1.13.3
+  - go: 1.15
 before_script:
 - mkdir -p bin
 - wget https://github.com/golang/dep/releases/download/v0.5.1/dep-linux-amd64 -O bin/dep


pohly:
I am going to submit update PRs in the various repos to see what breaks where. Those will be marked as WIP because of that and because this PR here must be merged as-is first.

pohly:
@jsafrane: can you perhaps lgtm? Most of the test PRs look good, only csi-test has one issue: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-test/282/pull-kubernetes-csi-csi-test/1293459960746741760

travis.yml
```diff
@@ -6,7 +6,7 @@ git:
   depth: false
 matrix:
   include:
-  - go: 1.13.3
+  - go: 1.15
```
msau42, line 9: We no longer use travis for builds right?

pohly: Perhaps - I wasn't sure whether it had been turned off everywhere.

As far as csi-release-tools is concerned, there might still be some repo which depends on TravisCI, therefore travis.yml is still present and defines the Go version also for Prow (as before):
https://github.com/kubernetes-csi/csi-release-tools/blob/d8a2530054bac287ba9a290f7add8a3423d56f4c/prow.sh#L96-L107

Issue #102, opened 2020-09-25
Canary jobs need to use the staging repo


msau42:
We now have prod images in k8s.gcr.io/sig-storage, but canary images are in gcr.io/k8s-staging-sig-storage

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-csi-canary-on-kubernetes-1-17/1309518998437629952/build-log.txt
```
  Warning  Failed     5m38s (x2 over 5m50s)  kubelet            Failed to pull image "k8s.gcr.io/sig-storage/csi-node-driver-registrar:canary": rpc error: code = NotFound desc = failed to pull and unpack image "k8s.gcr.io/sig-storage/csi-node-driver-registrar:canary": failed to resolve reference "k8s.gcr.io/sig-storage/csi-node-driver-registrar:canary": k8s.gcr.io/sig-storage/csi-node-driver-registrar:canary: not found
```

msau42:
/assign

Testing in https://github.com/kubernetes-csi/csi-driver-host-path/pull/202

Issue #86, opened 2020-05-19
Moving images away from quay.io?


xing-yang:
CI is failing because quay.io is down.  It's reported in this issue:
https://github.com/kubernetes/kubernetes/issues/91242

Should we move our images to a different repo, i.e., the community infra?

CC @msau42 @pohly

xing-yang:
One update from https://status.quay.io/ says that "The number of open connections to the Quay.io database increased significantly causing an outage of the service, we are still investigating the root cause of this spike in DB connections. "

xing-yang:
We have staging repos for CSI in GCR and the image promotion process.  It's used for node-driver-registrar only right now.  Should we do this for all sidecar repos?
https://github.com/kubernetes-csi/node-driver-registrar/blob/master/cloudbuild.yaml

justaugustus:
Thanks for opening this, @xing-yang!
I poked @hasheddan to get the conversation started in https://github.com/kubernetes/kubernetes/issues/91242 / https://kubernetes.slack.com/archives/CJH2GBF7Y/p1589887799251500.

Where possible, images that are used by that community should be on community infra.
I'd suggest moving anything currently hosted at quay.io/k8scsi be moved in a CSI staging repository and then promoted to our prod registry.

Let me know how I can help and feel free to assign me in reviews. :)

xing-yang:
Thanks @justaugustus!

pohly:
Work is in progress. It's not just about pushing to some other registry, but also about using an entirely different build infrastructure and producing images for multiple different architectures, ideally including Windows.

As a next step it would help if someone could approve https://github.com/kubernetes-csi/node-driver-registrar/pull/82. As I said in the review of that PR, we need to figure out how building of Windows images can be unified with the approach taken for all other platforms (i.e. build the binary inside the multi-stage Dockerfile, using the same base images if possible).

/cc @vitt-bagal @jingxu97 @ddebroy

pohly:
Previous PRs:
- https://github.com/kubernetes-csi/node-driver-registrar/pull/68
- https://github.com/kubernetes-csi/node-driver-registrar/pull/72

pohly:
Tentative plan:
- [ ] try out cloud build one last time after merging https://github.com/kubernetes-csi/node-driver-registrar/pull/84
- [ ] move common code and files into `csi-release-tools`
- [ ] update node-driver-registrar with that
[...]

PR #103, opened 2020-09-26
Use staging registry for canary tests

msau42:
Fixes: #102 

/kind failing-test

Testing in https://github.com/kubernetes-csi/csi-driver-host-path/pull/202


pohly (commented):
Looks good to me. However, we still need to ensure that the canary images actually are available there at all times. This is tracked as a TODO in https://github.com/kubernetes-csi/csi-release-tools/issues/86#issuecomment-634509901

msau42:
> Looks good to me. However, we still need to ensure that the canary images actually are available there at all times. This is tracked as a TODO in [#86 (comment)](https://github.com/kubernetes-csi/csi-release-tools/issues/86#issuecomment-634509901)

Yup, we'll need to take care of that in the next month or so

PR #108, opened 2020-10-19
Add go ldflags using LDFLAGS at the time of compilation

bnrjee:
What type of PR is this?
/kind bug

What this PR does / why we need it:
This PR allows adding ldflags while running make. We are trying to build binaries without debug info. With this change we will be able to pass -s flag as an argument to Makefile.

**Does this PR introduce a user-facing change?**:

```release-note
Add go ldflags using LDFLAGS at the time of compilation.
```

### DIFF
--- a/build.make
+++ b/build.make
@@ -69,12 +69,17 @@ endif
 # toolchain.
 BUILD_PLATFORMS =
 
+# Add go ldflags using LDFLAGS at the time of compilation.
+IMPORTPATH_LDFLAGS = -X main.version=$(REV) 
+EXT_LDFLAGS = -extldflags "-static"
+LDFLAGS = 
+FULL_LDFLAGS = $(LDFLAGS) $(IMPORTPATH_LDFLAGS) $(EXT_LDFLAGS)
 # This builds each command (= the sub-directories of ./cmd) for the target platform(s)
 # defined by BUILD_PLATFORMS.
 $(CMDS:%=build-%): build-%: check-go-version-go
 	mkdir -p bin
 	echo '$(BUILD_PLATFORMS)' | tr ';' '\n' | while read -r os arch suffix; do \
-		if ! (set -x; CGO_ENABLED=0 GOOS="$$os" GOARCH="$$arch" go build $(GOFLAGS_VENDOR) -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o "./bin/$*$$suffix" ./cmd/$*); then \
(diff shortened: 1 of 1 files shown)


bnrjee:
I have run some tests with external-attacher.

https://github.com/kubernetes-csi/external-attacher/pull/265#issuecomment-710302393

build.make
```diff
@@ -69,12 +69,15 @@ endif
 # toolchain.
 BUILD_PLATFORMS =
 
+# Add go ldflags using LDFLAGS at the time of compilation.
+LDFLAGS ?= -X main.version=$(REV) -extldflags "-static"
```
pohly, line 73: When using `make LDFLAGS=<something>`, the user has to remember to include these defaults in `<something>`. This looks a bit error-prone.

What about this?

```
# LDFLAGS can be set with "make LDFLAGS=..." to add custom link flags.
LDFLAGS =

# Some link flags are used automatically. This can be suppressed with "make FULL_LDFLAGS=..."
FULL_LDFLAGS =  -X main.version=$(REV) -extldflags "-static" $(LDFLAGS)

... -ldflags '$(FULL_LDFLAGS)' ...
```

bnrjee: @pohly, thanks for reviewing this. Will this suggested change allow adding extldflags ? The idea behind this PR is to make sure, we can add any ldflags we require.

bnrjee: I have added a few more variables to make sure everything is configurable.

build.make
```diff
@@ -69,12 +69,17 @@ endif
 # toolchain.
 BUILD_PLATFORMS =
 
+# Add go ldflags using LDFLAGS at the time of compilation.
+IMPORTPATH_LDFLAGS ?= -X main.version=$(REV) 
+EXT_LDFLAGS ?= -extldflags "-static"
+LDFLAGS = 
+FULL_LDFLAGS = $(LDFLAGS) $(IMPORTPATH_LDFLAGS) $(EXT_LDFLAGS)
```
pohly, line 76: This seems inconsistent: why are `IMPORTPATH_LDFLAGS` and `EXT_LDFLAGS` set with `?=` and the others with `=`? Was that intentional?

I think I prefer `=` because that then only allows overriding only via `make <VARNAME>=`. In particular, `LDFLAGS` might be something that a user has set in the environment when invoking make, without being aware that it affects the Go compilation - it's not likely, but I'd prefer to avoid surprises.

PR #110, opened 2020-10-27
prow.sh: work around "kind build node-image" failure

pohly:
Kind 0.9.0 accesses the "bazel-out" directory with a relative path,
which only works when the command is invoked inside the Kubernetes
source code
directory (https://github.com/kubernetes-sigs/kind/issues/1910).


pohly:
Tested in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_external-provisioner/512/pull-kubernetes-csi-external-provisioner-alpha-1-19-on-kubernetes-master/1321049239576383488

PR #111, opened 2020-11-12
Update snapshot CRD version

xing-yang:


xing-yang:
CI has passed on this PR: https://github.com/kubernetes-csi/csi-driver-host-path/pull/221

prow.sh
```diff
@@ -330,7 +330,14 @@ configvar CSI_PROW_E2E_ALPHA_GATES_LATEST 'GenericEphemeralVolume=true,CSIStorag
 configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_GATES "${csi_prow_kubernetes_version_suffix}")" "alpha E2E feature gates"
 
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
-configvar CSI_SNAPSHOTTER_VERSION 'v3.0.0' "external-snapshotter version tag"
+default_csi_snapshotter_version () {
+	if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" == "canary" ]; then
+		echo "master"
+	else
+		echo "v3.0.0"
```
msau42, line 337: should this be the latest patch?

Longer term, it would be good to find a way that doesn't require updating release-tools after every release.

prow.sh
```diff
@@ -330,7 +330,14 @@ configvar CSI_PROW_E2E_ALPHA_GATES_LATEST 'GenericEphemeralVolume=true,CSIStorag
 configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_GATES "${csi_prow_kubernetes_version_suffix}")" "alpha E2E feature gates"
 
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
-configvar CSI_SNAPSHOTTER_VERSION 'v3.0.0' "external-snapshotter version tag"
+default_csi_snapshotter_version () {
+	if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" == "canary" ]; then
```
pohly, line 334: `==` is an unnecessary bashism

prow.sh
xing-yang: v3.0.2?

xing-yang: Updated to v3.0.2

pohly: I proposed this logic here based on the observation that the hard-coded v3.0.0 was wrong for some jobs.

What I hadn't considered is that whether it would make sense to move this setting entirely into the Prow jobs. I think it actually does make sense, because it's at the same level as "let's run with this version of Kubernetes" and the API for it is fairly stable (the assumption is that also future releases will deploy via four URLs).

So the same logic can also go into https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/gen-jobs.sh such that the Prow job always sets `CSI_PROW_KUBERNETES_VERSION`, including the case where we currently use `v3.0.0`.

That opens the door for adding more test jobs for other combinations, like external-snapshotter v4.0.0 with controller from v3.0.0.

prow.sh
```diff
@@ -332,6 +332,10 @@ configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
 configvar CSI_SNAPSHOTTER_VERSION 'v3.0.0' "external-snapshotter version tag"
 
+# TODO(xyang): Replace with release build when available
+# Temporary external-snapshotter client tag to use for the snapshotter CRD
+configvar CSI_SNAPSHOTTER_CRD_VERSION 'v4.0.0-rc1' "external-snapshotter version tag"
```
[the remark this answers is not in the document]
xing-yang: This is an older version.  It is not there in the latest PR.

prow.sh
xing-yang: Fixed here: https://github.com/kubernetes-csi/csi-release-tools/pull/112

prow.sh
pohly: https://github.com/kubernetes/test-infra/pull/19939 is a PR which bumps the version through the Prow job configs.

PR #113, opened 2020-11-18
Fix the install of snapshot CRDs and controller

xing-yang:
This PR installs snapshot CRDs and rbac rules from the repo
and installs snapshot controller from a local image if it is
a PR in external-snapshotter repo.
Otherwise it uses main or a stable version.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -700,6 +700,10 @@ install_csi_driver () {
 install_snapshot_crds() {
   # Wait until volumesnapshot CRDs are in place.
   CRD_BASE_DIR="https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/${CSI_SNAPSHOTTER_VERSION}/client/config/crd"
+  if [[ ${REPO_DIR} == *"external-snapshotter"* ]]; then
+      CRD_BASE_DIR="${REPO_DIR}/client/config/crd"
+  fi
+  echo "Installing snapshot CRDs from ${CRD_BASE_DIR}"
   kubectl apply -f "${CRD_BASE_DIR}/snapshot.storage.k8s.io_volumesnapshotclasses.yaml" --validate=false
   kubectl apply -f "${CRD_BASE_DIR}/snapshot.storage.k8s.io_volumesnapshots.yaml" --validate=false
   kubectl apply -f "${CRD_BASE_DIR}/snapshot.storage.k8s.io_volumesnapshotcontents.yaml" --validate=false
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -733,8 +746,55 @@ install_snapshot_controller() {
     sleep 10   
   done
 
+  SNAPSHOT_CONTROLLER_YAML="${CONTROLLER_DIR}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+  if [[ ${REPO_DIR} == *"external-snapshotter"* ]]; then
+      NEW_TAG="csiprow"
```
msau42, line 751: Can you add a comment where you got this logic from?

xing-yang:
Submitted the following PRs to test this fix.  CI has passed on both PRs:
https://github.com/kubernetes-csi/external-snapshotter/pull/427
https://github.com/kubernetes-csi/csi-driver-host-path/pull/224

prow.sh
```diff
@@ -733,8 +746,55 @@ install_snapshot_controller() {
...
+      # shellcheck disable=SC2086
+      # Ignore: Use find instead of ls to better handle non-alphanumeric filenames.
+      # shellcheck disable=SC2012
+      for i in $(ls ${SNAPSHOT_CONTROLLER_YAML} | sort); do
+          echo "   $i"
+          # Ignore: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
+          # shellcheck disable=SC2002
+          # Ignore: See if you can use ${variable//search/replace} instead.
+          # shellcheck disable=SC2001
+          modified="$(cat "$i" | while IFS= read -r line; do
```
msau42, line 768: I think longer term we should consider using kustomize to patch the image so we don't need to maintain this logic.

msau42: https://github.com/kubernetes-csi/csi-release-tools/issues/114

prow.sh
xing-yang: This is here: https://github.com/kubernetes-csi/csi-release-tools/pull/113/files#diff-c2d96beb8cf73c62f7ae059aff96e381a8d73317df3ce356f36ba6b46338959dR1115

The image built from PR gets a "csiprow" tag.

prow.sh
xing-yang: Sure

prow.sh
xing-yang: Added comments.

xing-yang:
@msau42 addressed your comments.  PTAL.  Thanks.

prow.sh
msau42: Oh sorry, I was referring to the logic below where we're replacing the image tag

xing-yang: That's from deploy-hostpath.sh: https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/util/deploy-hostpath.sh#L155

xing-yang: Updated.

PR #116, opened 2020-11-24
allow export registry name

andyzhangx:


build.make
```diff
@@ -20,7 +20,7 @@
 
 # This is the default. It can be overridden in the main Makefile after
 # including build.make.
-REGISTRY_NAME=quay.io/k8scsi
+REGISTRY_NAME?=quay.io/k8scsi
```
pohly, line 23: This I understand: you want to do `REGISTRY_NAME=foo make push` instead of `make push REGISTRY_NAME=foo`.

build.make
```diff
@@ -51,7 +51,7 @@ IMAGE_TAGS+=$(shell git branch -r --points-at=HEAD | grep 'origin/release-' | gr
 IMAGE_TAGS+=$(shell tagged="$$(git describe --tags --match='v*' --abbrev=0)"; if [ "$$tagged" ] && [ "$$(git rev-list -n1 HEAD)" = "$$(git rev-list -n1 $$tagged)" ]; then echo $$tagged; fi)
 
 # Images are named after the command contained in them.
-IMAGE_NAME=$(REGISTRY_NAME)/$*
+IMAGE_NAME?=$(REGISTRY_NAME)/$*
```
pohly, line 54: But why this?

build.make
andyzhangx: correct

build.make
andyzhangx: mainly I want to:
```
export REGISTRY_NAME=...
export IMAGE_NAME=...
make build
make container
```

with this PR, it could tag image name with correct registry name, otherwise need workaround, thanks.

pohly: But why set `IMAGE_NAME`? It's a combination of other input variables. You shouldn't need to set it.

andyzhangx: I think there is var name conflict in NFS Makefile, reverted this change. PTAL.

pohly (approved):
/lgtm

Now it makes sense to me :grin:

PR #117, opened 2020-11-29
upgrade csi-test to v4.0.2

fengzixu:
### Reason

The current csi-test version doesn't support the new capabilities, like VOLUME_CONDITION and GET_VOLUME. Both of them are needed by the external-health-monitor project and other projects which are integrating with external-health-monitor: https://github.com/kubernetes-csi/csi-driver-host-path/pull/210.

In that case, I think we need to upgrade the version of csi-test

```release-note
upgrade version of csi-test to v4.0.2
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -227,7 +227,7 @@ configvar CSI_PROW_E2E_IMPORT_PATH "k8s.io/kubernetes" "E2E package"
 # of the cluster. The alternative would have been to (cross-)compile csi-sanity
 # and install it inside the cluster, which is not necessarily easier.
 configvar CSI_PROW_SANITY_REPO https://github.com/kubernetes-csi/csi-test "csi-test repo"
-configvar CSI_PROW_SANITY_VERSION 5421d9f3c37be3b95b241b44a094a3db11bee789 "csi-test version" # latest master
+configvar CSI_PROW_SANITY_VERSION af2d21f66c88d7b4956c17d8123cdc563ca0ba0a "csi-test version" # latest master
 configvar CSI_PROW_SANITY_IMPORT_PATH github.com/kubernetes-csi/csi-test "csi-test package"
 configvar CSI_PROW_SANITY_SERVICE "hostpath-service" "Kubernetes TCP service name that exposes csi.sock"
 configvar CSI_PROW_SANITY_POD "csi-hostpathplugin-0" "Kubernetes pod with CSI driver"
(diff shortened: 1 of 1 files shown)


xing-yang:
Hi @pohly, 

Volume health support was added to csi-test v4.0: https://github.com/kubernetes-csi/csi-test/blob/v4.0.2/CHANGELOG/CHANGELOG-4.0.md#other-notable-changes-1.

@fengzixu is working on adding volume health support to csi hostpath driver, but CI can't pass now because csi sanity test still uses an older version of csi-test: https://github.com/kubernetes-csi/csi-driver-host-path/pull/210/files#r530651591.

Can we update prow.sh to use csi-test v4.0.2?

prow.sh
```diff
@@ -227,8 +227,8 @@ configvar CSI_PROW_E2E_IMPORT_PATH "k8s.io/kubernetes" "E2E package"
 # of the cluster. The alternative would have been to (cross-)compile csi-sanity
 # and install it inside the cluster, which is not necessarily easier.
 configvar CSI_PROW_SANITY_REPO https://github.com/kubernetes-csi/csi-test "csi-test repo"
-configvar CSI_PROW_SANITY_VERSION 5421d9f3c37be3b95b241b44a094a3db11bee789 "csi-test version" # latest master
-configvar CSI_PROW_SANITY_IMPORT_PATH github.com/kubernetes-csi/csi-test "csi-test package"
+configvar CSI_PROW_SANITY_VERSION af2d21f66c88d7b4956c17d8123cdc563ca0ba0a "csi-test version" # latest master
```
pohly, line 230: A tag would be more readable and make the comment obsolete.

If I remember correctly, a hash was used only because there was no suitable release.

xing-yang: Makes sense.

@fengzixu, can you use tag v4.0.2 instead of a hash?

fengzixu: Done

xing-yang: Can you apply this change temporarily on your PR in csi-driver-host-path and see if it passes CI?  Unit test will fail but other CI should pass if this works.

fengzixu: Sure

fengzixu:
https://github.com/kubernetes-csi/csi-driver-host-path/pull/210 I think we can approve and merge this PR now.

pohly (commented):
/lgtm
/approve

PR #118, opened 2020-12-04
cloud build: double the timeout, now 1 hour

pohly:
Builds of external-snapshotter sometimes timed out because it builds
three images and that takes almost half an hour (the previous limit)
and more if the network is slow.

There's no good way to set timeouts per repo and a too high timeout
for shouldn't cause any issue, therefore the timeout gets increased
everywhere.

/cc @xing-yang


pohly:
Examples:
* 28min (success): https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/canary-external-snapshotter-push-images/1333532812074029056
* 30min (timeout): https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/canary-external-snapshotter-push-images/1330996023434153984

PR #121, opened 2020-12-24
Check namespace for snapshot-controller

kvaps:
This PR adds additional check for namespace specified in manifest for snapshot-controller
needed for https://github.com/kubernetes-csi/external-snapshotter/pull/459#issuecomment-750934403

```release-note
Check namespace for snapshot-controller tests
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -796,16 +796,17 @@ install_snapshot_controller() {
 	  echo "kubectl apply -f ${SNAPSHOT_CONTROLLER_YAML}(modified)"
       done
   else
-      echo "kubectl apply -f ${CONTROLLER_DIR}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
-      kubectl apply -f "${CONTROLLER_DIR}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml"
+      echo "kubectl apply -f $SNAPSHOT_CONTROLLER_YAML"
+      kubectl apply -f "$SNAPSHOT_CONTROLLER_YAML"
   fi
 
   cnt=0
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -801,11 +801,12 @@ install_snapshot_controller() {
   fi
 
   cnt=0
-  expected_running_pods=$(curl https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/"${CSI_SNAPSHOTTER_VERSION}"/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml | grep replicas | cut -d ':' -f 2-)
-  while [ "$(kubectl get pods -l app=snapshot-controller | grep 'Running' -c)" -lt "$expected_running_pods" ]; do
+  expected_running_pods=$(awk -F '[: ]+' '$1 == "" && $2 == "replicas" {print $3; exit}' "${CONTROLLER_DIR}/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml")
```
pohly, line 804: `CONTROLLER_DIR` is going to be a URL when testing something other than external-snapshotter. `awk` will fail in that case, won't it?

pohly (changes requested):
We need a test PR with this change in some repo other than external-snapshotter.

prow.sh
kvaps: Most possible yes, if target repository has no `setup-snapshot-controller.yaml` then it be failed.
We can add additional logic, to check if repository has this file, if not, it will be applied from the URL.

xing-yang:
@kvaps can you submit a test PR with this change in https://github.com/kubernetes-csi/csi-driver-host-path?

kvaps:
Sure, job is done kubernetes-csi/csi-driver-host-path#235

pohly (commented):
/lgtm
/approve

PR #124, opened 2021-01-07
Fix spelling and boilerplate errors

### LINKED ISSUE #123: Fix spelling and boilerplate errors
Due to spelling and boilerplate errors in some files, spellcheck and boilerplate tests are failing.

Boilerplates are wrong in `cloudbuild.sh`, `.prow.sh`, `verify-subtree.sh` and `prow.sh`.
Some spellings are wrong in `SIDECAR_RELEASE_PROCESS.md` and `prow.sh`.

sachinkumarsingh092:
Some spelling and boilerplate errors are fixed which caused the failure of these tests.

Fixes #123

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,7 +1,23 @@
 #! /bin/bash -e
+
+# Copyright 2021 The Kubernetes Authors.
 #
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # This is for testing csi-release-tools itself in Prow. All other
 # repos use prow.sh for that, but as csi-release-tools isn't a normal
 # repo with some Go code in it, it has a custom Prow test script.
 
 ./verify-shellcheck.sh "$(pwd)"
+./verify-spelling.sh "$(pwd)"
+./verify-boilerplate.sh "$(pwd)"
--- a/cloudbuild.sh
+++ b/cloudbuild.sh
@@ -1,5 +1,19 @@
 #! /bin/bash
 
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # shellcheck disable=SC1091
 . release-tools/prow.sh
 
--- a/verify-spelling.sh
+++ b/verify-spelling.sh
@@ -20,9 +20,11 @@ set -o pipefail
 
 TOOL_VERSION="v0.3.4"
 
-# cd to the root path
-ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
-cd "${ROOT}"
+# The csi-release-tools directory.
+TOOLS="$(dirname "${BASH_SOURCE[0]}")"
+
+# Directory to check. Default is the parent of the tools themselves.
+ROOT="${1:-${TOOLS}/..}"
 
 # create a temporary directory
 TMP_DIR=$(mktemp -d)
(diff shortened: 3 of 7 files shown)


.prow.sh
```diff
@@ -1,5 +1,19 @@
 #! /bin/bash -e
+
+# Copyright 2014 The Kubernetes Authors.
```
xing-yang, line 3: Why is this 2014?

sachinkumarsingh092: Shall I change it to 2019 (sam as in `prow.sh`)?

.prow.sh
```diff
@@ -1,5 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # This is for testing csi-release-tools itself in Prow. All other
 # repos use prow.sh for that, but as csi-release-tools isn't a normal
 # repo with some Go code in it, it has a custom Prow test script.
```
msau42, line 19: Please add the new tests to this file so they will run during pull jobs.

sachinkumarsingh092: Shall I also change the date (as reviewed by @xing-yang above) in the boilerplate to 2019?

msau42: Yes, please change the year 2021

sachinkumarsingh092: Done!

cloudbuild.sh
```diff
@@ -1,5 +1,19 @@
 #! /bin/bash
 
+# Copyright 2014 The Kubernetes Authors.
```
andyzhangx, line 3: 2021

.prow.sh
andyzhangx: should use 2021

cloudbuild.sh
sachinkumarsingh092: Done!

andyzhangx (approved):
/lgtm
/approve

verify-spelling.sh
```diff
@@ -21,7 +21,7 @@ set -o pipefail
 TOOL_VERSION="v0.3.4"
 
 # cd to the root path
-ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
```
pohly, line 24: The commit message only mentions "Some spelling and boilerplate errors are fixed which caused the failing of these tests." but this here is a functional change.

What is "the root path" - root of the project or root of the "csi-release-tools" scripts? I suspect that with change, only files in the "release-tools" directory will be verified...

Can `ROOT` and the associated `cd` perhaps be removed such that `verify-spelling.sh` will check files where it is invoked?

sachinkumarsingh092: We need `ROOT` and `cd` to get to the top-level directory. There was an error with this one so I fixed it using `git rev-parse --show-toplevel` which always points to the absolute path of top-level directory.

pohly: This seems unnecessarily complex and inflexible. It almost broke if I hadn't noticed this change.

Why not do this as in `verify-shellcheck.sh` where the first parameter determines which directory is to be checked?

pohly: `verify-boilerplate.sh` has the same bug: https://github.com/kubernetes-csi/csi-release-tools/blob/a8c35dc270229a1d8b8bb01b415119eb155ed0cf/verify-boilerplate.sh#L28-L33

sachinkumarsingh092: It happed before (and again) that when I use style similar to `verify-shellcheck.sh`, the tests fail both locally and in prow jobs. The error being, that when a `cd` for `ROOT` is used using the `verify-shellcheck.sh` or similar style, it changes directories into a non-git directory.
Can you please suggest a fix?

pohly: I don't follow. If `.prow.sh` and `prow.sh` both call the check scripts with `$(pwd)` as first parameter (my proposal), why should a "cd" command then land in a non-git directory?

.prow.sh
```diff
@@ -1,7 +1,23 @@
...
 
 ./verify-shellcheck.sh "$(pwd)"
+./verify-spelling.sh
+./verify-boilerplate.sh
```
pohly, line 23: You have to add `"$(pwd)"` here to override the default.

verify-spelling.sh
pohly: I missed that you updated the scripts. What you didn't update is `.prow.sh` - see my other comment.

verify-spelling.sh
```diff
@@ -20,8 +20,13 @@ set -o pipefail
...
 TOOL_VERSION="v0.3.4"
 
+# The csi-release-tools directory.
+TOOLS="$(dirname "${BASH_SOURCE[0]}")"
+
+# Directory to check. Default is the parent of the tools themselves.
+ROOT="${1:-${TOOLS}/..}"
+
 # cd to the root path
-ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
```
pohly, line 24: This cd is redundant. There is another `cd "${ROOT}"` further down before the actual check.

.prow.sh
sachinkumarsingh092: Done! Thanks

verify-spelling.sh
sachinkumarsingh092: Now I've fixed this with the style similar to that in `verify-shellcheck.sh`.

verify-spelling.sh
sachinkumarsingh092: Done!

verify-spelling.sh
pohly: My point was that the following lines can be removed completely:
```
# cd to the root path
cd "${ROOT}"
```

pohly: For the sake of clarity it would be good to move the "cd" below into the "check spelling" section:

```
# check spelling
RES=0
echo "Checking spelling..."
ERROR_LOG="${TMP_DIR}/errors.log"
cd "$ROOT"
```

Then it is obvious that `git ls-files` is meant to operate on files in that directory.

sachinkumarsingh092: Sorry, I misunderstood! It's fixed now.

pohly (approved):
/lgtm

PR #127, opened 2021-01-17
verify-boilerplate.sh: fix path to script

pohly:
The previous path was wrong because it was relative to the directory
being checked, not the release-tools directory:

```
csi-driver-host-path$ ./release-tools/verify-boilerplate.sh
Verifying boilerplate
./release-tools/verify-boilerplate.sh: line 37: /nvme/gopath/src/github.com/kubernetes-csi/csi-driver-host-path/boilerplate/boilerplate.py: No such file or directory
```

### DIFF
--- a/verify-boilerplate.sh
+++ b/verify-boilerplate.sh
@@ -31,8 +31,7 @@ TOOLS="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
 # Directory to check. Default is the parent of the tools themselves.
 ROOT="${1:-${TOOLS}/..}"
 
-boilerDir="${ROOT}/boilerplate"
-boiler="${boilerDir}/boilerplate.py"
+boiler="${TOOLS}/boilerplate/boilerplate.py"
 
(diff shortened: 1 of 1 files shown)


pohly:
This might have been missed because the test succeeds when that happens because the return code of the script is not checked.

verify-boilerplate.sh
```diff
@@ -31,8 +31,7 @@ TOOLS="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
 # Directory to check. Default is the parent of the tools themselves.
 ROOT="${1:-${TOOLS}/..}"
 
-boilerDir="${ROOT}/boilerplate"
-boiler="${boilerDir}/boilerplate.py"
+boiler="${TOOLS}/boilerplate/boilerplate.py"
```
msau42, line 34: should we also fix the script to check the return code?

pohly: Would be better, yes. I don't have a solution readily available, though.

PR #128, opened 2021-01-19
fix: fix a bug of building csi-sanity

fengzixu:
```release-note
None
```

For two issues:
1. there is no test file under /csi-test/cmd/csi-sanity in v4.0.2 version. It was old code in v2.2.0. In that case, we should use go test -c to build binary file of csi-sanity https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/release-tools/prow.sh#L883
2. fetch csi-test from upstream always

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -228,7 +228,7 @@ configvar CSI_PROW_E2E_IMPORT_PATH "k8s.io/kubernetes" "E2E package"
 # and install it inside the cluster, which is not necessarily easier.
 configvar CSI_PROW_SANITY_REPO https://github.com/kubernetes-csi/csi-test "csi-test repo"
 configvar CSI_PROW_SANITY_VERSION v4.0.2 "csi-test version" # v4.0.2
-configvar CSI_PROW_SANITY_IMPORT_PATH github.com/kubernetes-csi/csi-test/v4 "csi-test package"
+configvar CSI_PROW_SANITY_PACKAGE_PATH github.com/kubernetes-csi/csi-test "csi-test package"
 configvar CSI_PROW_SANITY_SERVICE "hostpath-service" "Kubernetes TCP service name that exposes csi.sock"
 configvar CSI_PROW_SANITY_POD "csi-hostpathplugin-0" "Kubernetes pod with CSI driver"
 configvar CSI_PROW_SANITY_CONTAINER "hostpath" "Kubernetes container with CSI driver"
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -879,8 +879,10 @@ install_sanity () (
         return
     fi
 
-    git_checkout "${CSI_PROW_SANITY_REPO}" "${GOPATH}/src/${CSI_PROW_SANITY_IMPORT_PATH}" "${CSI_PROW_SANITY_VERSION}" --depth=1 || die "checking out csi-sanity failed"
-    run_with_go "${CSI_PROW_GO_VERSION_SANITY}" go test -c -o "${CSI_PROW_WORK}/csi-sanity" "${CSI_PROW_SANITY_IMPORT_PATH}/cmd/csi-sanity" || die "building csi-sanity failed"
+    git_checkout "${CSI_PROW_SANITY_REPO}" "${GOPATH}/src/${CSI_PROW_SANITY_PACKAGE_PATH}" "${CSI_PROW_SANITY_VERSION}" --depth=1 || die "checking out csi-sanity failed"
+    cd "${GOPATH}/src/${CSI_PROW_SANITY_PACKAGE_PATH}/cmd/csi-sanity" || die "enter csi-sanity directory failed"
+    go build || die "building csi-sanity failed"
+    mv ./csi-sanity "${CSI_PROW_WORK}"
```
pohly, line 885: Don't call `go` directly, always wrap it with `run_with_go` to ensure that the right version of Go is in the `PATH`. Permanently changing the directory can also be confusing. Better use:
```
git_checkout "${CSI_PROW_SANITY_REPO}" "${GOPATH}/src/${CSI_PROW_SANITY_PACKAGE_PATH}" "${CSI_PROW_SANITY_VERSION}" --depth=1 || die "checking out csi-sanity failed"
( cd "${GOPATH}/src/${CSI_PROW_SANITY_PACKAGE_PATH}/cmd/csi-sanity" && run_with_go "${CSI_PROW_GO_VERSION_SANITY}" go build -o "${CSI_PROW_WORK}/csi-sanity" ./cmd/csi-sanity ) || die "building csi-sanity failed"
```

pohly (changes requested):

prow.sh
fengzixu: Fixed

fengzixu:
@pohly Please review and merge this PR first. After that, I will fix error that was caused by release-tool in  https://github.com/kubernetes-csi/csi-driver-host-path/pull/234 and https://github.com/kubernetes-csi/csi-driver-host-path/pull/210

prow.sh
fengzixu:
```
Wed Jan 20 03:46:22 UTC 2021 go1.15.5 /home/prow/go/src/github.com/kubernetes-csi/csi-test/cmd/csi-sanity$ go build -o /home/prow/go/pkg/csiprow.Fc5cBsm6I0/csi-sanity ./cmd/csi-sanity
stat /home/prow/go/src/github.com/kubernetes-csi/csi-test/cmd/csi-sanity/cmd/csi-sanity: directory not found
ERROR: building csi-sanity failed
```

Maybe `( cd "${GOPATH}/src/${CSI_PROW_SANITY_PACKAGE_PATH}/cmd/csi-sanity" && run_with_go "${CSI_PROW_GO_VERSION_SANITY}" go build -o "${CSI_PROW_WORK}/csi-sanity") || die "building csi-sanity failed"` is correct command

pohly: Either changing into `cmd/csi-sanity` and `go build` without parameters or changing into the root and then using `go build ./cmd/csi-sanity` work, but not doing both :grin: 

I don't have a preference either way and verified that that the current commit works, so let's use that.

/lgtm
/approve

Issue #7, opened 2019-02-18
automatically update vendored dependencies


pohly:
In our various Kubernetes-CSI repos we should regularly update dependencies to ensure that we have included all upstream bug fixes.

This includes:
- running `dep ensure -update`
- running `git subtree pull` for `release-tools` (see https://github.com/kubernetes-csi/csi-release-tools/blob/master/README.md#sharing-and-updating)
- creating a new PR for changes (if none is pending currently) or updating an existing PR by force-pushing to the branch

Then once the PR is pending, CI testing will cover the initial round of testing. Even if that fails, having a PR pending is a reminder to the maintainer that the component is not up-to-date.

The steps above could be automated with a script, like `update-dependencies.sh` in this repo. The `hub` tool might be useful for managing the PRs. Bonus points for making it possible to update release branches.

Running this script for one or more component is probably best done by the same person at a regular cadence. If multiple people do it, we end up with multiple competing PRs. We could also try to move it into a Prow job, but that would be a second step.

@msau42 does that sound useful?

msau42:
Yes I think that sounds valuable.  It would be great if there was also a wrapper script that can update and create prs across all the branches

pohly:
Quick-and-dirty script that I just used to update csi-relase-tools:
```
#!/bin/sh -x

die () {
    echo >&2 "\nERROR: $@\n"
    exit 1
}

while read repo branches; do
    if [ "$repo" != "#" ]; then
    (
        cd $repo || die "$repo: does not exit"
        git fetch origin || die "$repo: git fetch"
        for i in $branches; do
            git checkout -B prow-update-$i origin/$i || die "$repo:$i checkout"
            git subtree pull --prefix=release-tools -m "release-tools: update" https://github.com/kubernetes-csi/csi-release-tools.git master || die "$repo:$i update release tools"
            changes=$(git log --no-merges --oneline origin/$i..prow-update-$i)
            if [ "$changes" ]; then
                make test || die "$repo:$i make test"
                git push pohly prow-update-$i || die "pohly:prow-update-$i push failed - probably there is already an unmerged branch and pending PR"
                hub pull-request -F - -b kubernetes-csi/$repo:$i -h pohly/$repo:prow-update-$i <<EOF
$i: update release-tools

Commit summary:
$changes

\`\`\`release-note
NONE
\`\`\`
EOF
                if [ "$?" -ne 0 ]; then
                    die "$repo:$i pull-request"
                fi
            fi
[...]

PR #129, opened 2021-01-19
README.md: document usage of --squash

pohly:


README.md
```diff
@@ -44,10 +44,12 @@ is the recommended way of maintaining a copy of the rules inside the
 changes also locally, test them and then push them back to the shared
 repository at a later time.
 
+We no longer care about importing the full commit history, so `--squash` should be used.
```
xing-yang, line 47: We should still add those commits being squashed in the PR description, right?

pohly: That would be nice, but cannot be done easily unless developers start using the script instead of following these manual instructions.

pohly: Volunteers are welcome for working on integration of that script into this repo... see https://github.com/kubernetes-csi/csi-release-tools/issues/7

xing-yang: Can you add a line here to ask developers to copy the automatically generated commit message (which lists all the commits) to the PR description?

msau42: We need to be careful that these individual commit messages are not of the format "Merge pull request". It seems to be messing up the release notes tool and causes the same PR numbers from the sidecar repos to be picked up. https://github.com/kubernetes-csi/csi-driver-host-path/pull/238#discussion_r565535893

pohly: @msau42 I think the csi-driver-host-path 1.5.0 release note generation got confused because there were still some `release-tools` updates in the history since 1.4.0 which were done without `--squash`, for example: https://github.com/kubernetes-csi/csi-driver-host-path/commit/e53f3e85b7819546d2b66f32d5d7d4e6dc1020bc

Or do you have an example where that is still a problem when using squashing exclusively?

pohly: > Can you add a line here to ask developers to copy the automatically generated commit message (which lists all the commits) to the PR description?

I've extended the instructions, please take another look.

msau42: I don't have an example, but it's something we should watch out for when doing future release notes. I'm not sure how strict the release-notes parsing is.

xing-yang: It didn't cause problem for me when I generated release notes.  We probably can't wrap them with \`\`\`.

Issue #130, opened 2021-01-20
pre-merge jobs for testing usage of csi-release-tools in other components


pohly:
When a change is proposed for csi-release-tools, we need to be sure that the change works for the repos which later will get updated with the modified csi-release-tools. Ideally, updating csi-release-tools should work without changes in those repose, but that's not always possible.

Currently we rely on manual testing in other repos or test PRs in the other repos with the proposed change. Both is cumbersome to set up.

It would be useful to have Prow jobs for a number of key components (csi-driver-host-path, external-provisioner, external-snapshotter?) that automatically run some of the tests for that component (unit testing, sanity, E2E for one Kubernetes release?) with the updated csi-release-tools.

PR #131, opened 2021-01-26
prow.sh: support building Kubernetes for a specific version

pohly:
Checking out the Kubernetes source code only supported
branches (master, release-1.20) but not a version (like 1.20.0)
because tags need a leading "v". "git clone --branch" (despite the
name) is able to fetch a tag.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -65,6 +65,18 @@ get_versioned_variable () {
     echo "$value"
 }
 
+# This takes a version string like CSI_PROW_KUBERNETES_VERSION and
+# maps it to the corresponding git tag, branch or commit.
+version_to_git () {
+    version="$1"
+    shift
+    case "$version" in
+        latest) echo "master";;
+        release-*) echo "$version";;
+        *) echo "v$version";;
+    esac
+}
+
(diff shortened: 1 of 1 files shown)


pohly:
/assign @msau42 

This fixes Kubernetes 1.20 jobs, test PR is in https://github.com/kubernetes-csi/csi-driver-host-path/pull/236

The problem will be the older release branches. We have to update csi-release-tools on master *and* all `release-*` branches that are still maintained. Just knowing which those are is already a bit difficult (in particular when doing it automatically), hence my question in https://github.com/kubernetes-csi/docs/pull/407#discussion_r558254186.

The next problem then will be that for very old release branches, updating csi-release-tools to the latest master release may bring in also undesirable and/or incompatible changes. My original intention for csi-release-tools was to avoid forking it by keeping changes to it backwards-compatible - I have no idea how well that worked :cry: 

In general, keeping csi-release-tools working for various repos has been a bit challenging; see https://github.com/kubernetes-csi/csi-release-tools/issues/130 for an idea how that might be simplified with more automatic testing.

My proposal is:
- let's merge this PR
- update master in all repos
- update the latest release branch in all repos

PR #132, opened 2021-01-28
fix: default boilerplate path

yiyang5055:
add csi-release-tools to the program and do verify boilerplate does not work, because default boilerplate dir error.

### DIFF
--- a/boilerplate/boilerplate.py
+++ b/boilerplate/boilerplate.py
@@ -19,8 +19,6 @@ from __future__ import print_function
 import argparse
 import difflib
 import glob
-import json
-import mmap
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


pohly:
/ok-to-test

Do you have a test PR in some other repo with this commit?

boilerplate/boilerplate.py
```diff
@@ -33,12 +33,12 @@
...
 
 # Rootdir defaults to the directory **above** the repo-infra dir.
-rootdir = os.path.dirname(__file__) + "./../../../"
+rootdir = os.path.dirname(__file__) + "./../../"
 rootdir = os.path.abspath(rootdir)
 parser.add_argument(
     "--rootdir", default=rootdir, help="root directory to examine")
 
-default_boilerplate_dir = os.path.join(rootdir, "csi-driver-nfs/hack/boilerplate")
+default_boilerplate_dir = os.path.join(rootdir, "hack/boilerplate")
```
pohly, line 41: This is broken. If `rootdir` is the directory above "release-tools", then there is no "hack/boilerplate" directory in it.

I think `os.path.abspath(os.path.dirname(__file__))` is the right default?

pohly (changes requested):

boilerplate/boilerplate.py
yiyang5055: sorry, `hack/boilerplate` is my project directory.
yes, `default_boilerplate_dir=os.path.abspath(os.path.dirname(__file__))` is right

yiyang5055:
> /ok-to-test
> 
> Do you have a test PR in some other repo with this commit?

I tested it in my project after I modified it

pohly:
I verified this change in csi-driver-host-path.

/lgtm
/approve

PR #133, opened 2021-02-01
prow.sh: fix "on-master" prow jobs

pohly:
The recent introduction of version_to_git broke Prow jobs with
CSI_PROW_KUBERNETES_VERSION=latest because that version was first
converted to "master" and then once more to "vmaster". The second call
must be idempotent because "master" is already a branch name.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -71,7 +71,7 @@ version_to_git () {
     version="$1"
     shift
     case "$version" in
-        latest) echo "master";;
+        latest|master) echo "master";;
         release-*) echo "$version";;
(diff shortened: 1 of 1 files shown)


pohly:
One example that failed: https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-csi-1-20-on-kubernetes-master/1356210032831631360/build-log.txt

pohly:
Tested successfully (= fails without this, works with it) in https://github.com/kubernetes-csi/csi-driver-host-path/pull/242

/assign @msau42

PR #141, opened 2021-04-23
go-get-kubernetes.sh: make replace statement pruning optional

pohly:
Not pruning is safer because it allows adding k/k to a project at a
later time. Therefore adding all replace statements is the default.

A new flag can be used to clean up if desired. It's just a cosmetic
change because unused replace statements are ignored by Go.

/assign @jsafrane 
/cc @msau42

### DIFF
--- a/go-get-kubernetes.sh
+++ b/go-get-kubernetes.sh
@@ -26,14 +26,43 @@ set -o pipefail
 cmd=$0
 
 function help () {
-    echo "$cmd <kubernetes version = x.y.z> - update all components from kubernetes/kubernetes to that version"
+    cat <<EOF
+$cmd -p <kubernetes version = x.y.z>
+
+Update all components from kubernetes/kubernetes to that version.
+
+By default, replace statements are added for all Kubernetes packages,
+whether they are used or not.  This is useful when preparing a
+repository for using k8s.io/kubernetes, because those replace
+statements are needed to avoid "unknown revision v0.0.0" errors
+(https://github.com/kubernetes/kubernetes/issues/79384).
+
+With the optional -p flag, all unused replace statements are
+pruned. This makes go.mod smaller, but isn't required.
+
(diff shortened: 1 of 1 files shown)


msau42:
It's generally discouraged to import k/k packages directly.

pohly:
I know, but sometimes it cannot be avoided.

CSI driver developers who want to implement their own E2E test suite and reuse existing support code for that (k/k/test/e2e/framework, the storage test suites) have no choice, they have to import k/k. The effort to move the framework into staging has been abandoned and for the storage test suite that never has been the goal.

We don't have to make it harder by removing features that already existed.

jsafrane:
I checked output with `-p` and without it, it works as described in the help text and does not make go.mod worse.

/lgtm

sunnylovestiramisu:
I tried to run the script for 1.26.0 
```
./release-tools/go-get-kubernetes.sh -p 1.26.0-rc.0
```

But it failed with:
```
go: inconsistent vendoring in {package path}/external-attacher:
	k8s.io/api: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	k8s.io/apimachinery: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	k8s.io/client-go: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	k8s.io/component-base: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	k8s.io/csi-translation-lib: is replaced in go.mod, but not marked as replaced in vendor/modules.txt

	To ignore the vendor directory, use -mod=readonly or -mod=mod.
	To sync the vendor directory, run:
		go mod vendor

```

ERROR: could not obtain package list, both of these commands failed:

msau42:
Are you synced up to a recent version of external-attacher? The replace statements were removed in https://github.com/kubernetes-csi/external-attacher/pull/385

PR #144, opened 2021-04-28
pull-test.sh: test importing csi-release-tools into other repo

pohly:
This will be used by special Prow pull
jobs (https://github.com/kubernetes/test-infra/pull/21995) to ensure
that an update of csi-release-tools works in consuming repos.

### DIFF
--- a/pull-test.sh
+++ b/pull-test.sh
@@ -1,0 +1,32 @@
+#! /bin/sh
+
+# Copyright 2021 The Kubernetes Authors.
+#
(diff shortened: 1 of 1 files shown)


pohly:
/assign @msau42 

This should be working now.

One fix for the jobs is in https://github.com/kubernetes/test-infra/pull/22015 but is not needed as long as the tests pass, which they do at the moment - it's just not something that can be required because sometimes breaking changes must be merged into csi-release-tools first.

PR #147, opened 2021-05-06
Mock testing

pohly:
We already run the e2e.test binary from Kubernetes in different
invocations (serial, parallel, with and without alpha
features). Enabling also the in-tree mock tests is therefore easy and
makes sense in particular for our canary jobs where we try out how our
canary images work in existing deployments before updating those
deployments.

To role this out:
- merge this PR
- update csi-driver-host-path

We don't need to update jobs because the default is to include mock tests when
canary images are enabled.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -354,11 +354,7 @@ configvar CSI_SNAPSHOTTER_VERSION "$(default_csi_snapshotter_version)" "external
 # whether they can run with the current cluster provider, but until
 # they are, we filter them out by name. Like the other test selection
 # variables, this is again a space separated list of regular expressions.
-#
-# "different node" test skips can be removed once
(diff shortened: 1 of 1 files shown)


pohly:
I tried that out locally and it seemed to do the right thing (tests ran as intended). I had some random test flakes, but those went away when invoking tests with less parallelism, so it might be related to my environment. We won't know how well it works in Prow until it gets merged, though.

pohly:
Yeah! The new "pull-kubernetes-csi-*" jobs are already paying of: I hadn't tested the case without canary images, and that configuration got broken by this change. Should be fixed now.

pohly:
And now we need it in csi-driver-host-path: https://github.com/kubernetes-csi/csi-driver-host-path/pull/287

PR #148, opened 2021-05-07
prow.sh: enable -csi.checkpathcmd option in csi-sanity

dobsonj:
This enables the -csi.checkpathcmd option, which is needed in order
to test if TargetPath was successfully removed by the CSI driver as
part of NodeUnpublishVolume.

This is a follow up to https://github.com/kubernetes-csi/csi-test/pull/338
cc: @gnufied @pohly

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -975,6 +975,24 @@ if ! kubectl exec "${CSI_PROW_SANITY_POD}" -c "${CSI_PROW_SANITY_CONTAINER}" -- 
     exit 1
 fi
 EOF
+
+    cat >"${CSI_PROW_WORK}/checkdir_in_pod.sh" <<EOF
+#!/bin/sh
+CHECK_PATH=\$(cat <<SCRIPT
+if [ -f "\$@" ]; then
+    echo "file"
+elif [ -d "\$@" ]; then
+    echo "directory"
+elif [ -e "\$@" ]; then
+    echo "other"
+else
+    echo "not_found"
+fi
+SCRIPT
+)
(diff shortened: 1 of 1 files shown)


dobsonj:
FWIW, I tested this change under `csi-test/release-tools` and verified that the new test runs and passes there.
```
Node Service NodeUnpublishVolume
  should remove target path
  /root/go/src/csi-test/pkg/sanity/node.go:453
STEP: reusing connection to CSI driver at dns:///172.18.0.4:30931
STEP: creating mount and staging directories
STEP: creating a single node writer volume for expansion
STEP: Getting a node id
STEP: Staging and publishing a volume
STEP: node staging volume
STEP: publishing the volume on a node
STEP: Checking the target path exists
STEP: Unpublishing the volume
STEP: Checking the target path was removed
•
------------------------------
```

pohly:
/ok-to-test

This now should be tested with the csi-driver-host-path repo...

pohly:
Nope, I didn't add that driver repo. Might be worth adding.

pohly:
Looks like testing was indeed insufficient: it failed when updating csi-release-tools in csi-driver-host-path because the new flag was used with the old csi-test that didn't support it (https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-driver-host-path/295/pull-kubernetes-csi-csi-driver-host-path-1-21-test-on-kubernetes-1-21/1395031426440630272/build-log.txt).

Fix in https://github.com/kubernetes-csi/csi-release-tools/pull/152

PR #152, opened 2021-05-19
prow.sh: use csi-test v4.2.0

pohly:
The recently added -csi.checkpathcmd and the associated test are
supported only by csi-test v4.2.0.


pohly:
/hold

Let's test with https://github.com/kubernetes/test-infra/pull/22248 once it is available.

PR #159, opened 2021-05-25
prow.sh: more flexible CSI_PROW_DEPLOYMENT, part II

pohly:
Commit 61538bb776c5cae63d8dc40c68e6830389774c62 worked for some
jobs, but not for all, for example not for
CSI_PROW_KUBERNETES_VERSION=latest CSI_PROW_DEPLOYMENT=kubernetes-1.21

The approach taken now is to have one code path for all file checks
regardless whether CSI_PROW_DEPLOYMENT is set or not. The first
version number in the deployment name gets replaced with "latest" for
the fallback.

/assign @xing-yang


pohly:
This time I have tested several different job configurations locally...

But the ultimate test will only come after copying this into csi-driver-host-path.

PR #166, opened 2021-08-03
Explicitly reference PULL_BASE_REF as an env var in `make push-multiarch`

mauriciopoppe:
The kubernetes-csi/livenessprobe postsubmit prowjob https://k8s-testgrid.appspot.com/sig-storage-image-build#post-livenessprobe-push-images is failing, I checked in the [logs](https://storage.googleapis.com/kubernetes-jenkins/logs/post-livenessprobe-push-images/1420453703948701696/build-log.txt) that the makefile var `PULL_BASE_REF` isn't defined but it's used here:

https://github.com/kubernetes-csi/csi-release-tools/blob/ac8108f12f64eb61f489aaf7df611d93e77a1332/build.make#L188-L201

Probably the intention was to refer to the env var instead, in any case this PR references it as an env var.

Update: Michelle pointed out that https://k8s-testgrid.appspot.com/sig-storage-image-build#post-livenessprobe-push-images succeeded before this PR got merged, it looks like PULL_BASE_REF was set from to outside to an empty value

```release-note
NONE
```

/cc @pohly @msau42

### DIFF
--- a/build.make
+++ b/build.make
@@ -185,24 +185,24 @@ $(CMDS:%=push-multiarch-%): push-multiarch-%: check-pull-base-ref build-%
 		done; \
 		docker manifest push -p $(IMAGE_NAME):$$tag; \
 	}; \
-	if [ $(PULL_BASE_REF) = "master" ]; then \
+	if [ ${PULL_BASE_REF} = "master" ]; then \
 			: "creating or overwriting canary image"; \
 			pushMultiArch canary; \
-	elif echo $(PULL_BASE_REF) | grep -q -e 'release-*' ; then \
+	elif echo ${PULL_BASE_REF} | grep -q -e 'release-*' ; then \
 			: "creating or overwriting canary image for release branch"; \
-			release_canary_tag=$$(echo $(PULL_BASE_REF) | cut -f2 -d '-')-canary; \
+			release_canary_tag=$$(echo ${PULL_BASE_REF} | cut -f2 -d '-')-canary; \
 			pushMultiArch $$release_canary_tag; \
-	elif docker pull $(IMAGE_NAME):$(PULL_BASE_REF) 2>&1 | tee /dev/stderr | grep -q "manifest for $(IMAGE_NAME):$(PULL_BASE_REF) not found"; then \
+	elif docker pull $(IMAGE_NAME):${PULL_BASE_REF} 2>&1 | tee /dev/stderr | grep -q "manifest for $(IMAGE_NAME):${PULL_BASE_REF} not found"; then \
 			: "creating release image"; \
-			pushMultiArch $(PULL_BASE_REF); \
+			pushMultiArch ${PULL_BASE_REF}; \
 	else \
-			: "ERROR: release image $(IMAGE_NAME):$(PULL_BASE_REF) already exists: a new tag is required!"; \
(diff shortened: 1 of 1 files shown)


msau42:
It's passing now?
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-livenessprobe-push-images/1422340397211848704

mauriciopoppe:
The logs for this build (it shows as successful) https://storage.googleapis.com/kubernetes-jenkins/logs/post-livenessprobe-push-images/1422340397211848704/artifacts/build.log.

It's kinda puzzling that for the error shown in https://storage.googleapis.com/kubernetes-jenkins/logs/post-livenessprobe-push-images/1420453703948701696/build-log.txt there was no value for this env var

mauriciopoppe:
Even though it looks like the value is being taken from the env var, I believe it'd be better if this is shown explicitly in the command used and therefore I think we should still merge this PR. Another option would be to print the envvars at the start of the Makefile

build.make
```diff
@@ -72,6 +72,11 @@ endif
 # toolchain.
 BUILD_PLATFORMS =
 
+# PULL_BASE_REF will contain the base ref that was pushed to - for instance,
+# master or release-0.2 for a PR merge, or v0.2 for a tag.
+# See: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md
+PULL_BASE_REF =
```
pohly, line 78: "will contain the base ref" is a bit misleading. It only is non-empty if set explicitly when invoking make with `make PULL_BASE_REF=...`, which is not always the case.

`PULL_BASE_REF=... make` leads to an empty `PULL_BASE_REF` make variable. It doesn't matter at the moment because only `push-multiarch` depends on the variable but I find that a bit dangerous because it might bite us in the future when that changes.

pohly:
If the goal is to clarify where `PULL_BASE_REF` comes from, then we could use `${PULL_BASE_REF}` instead of `$(PULL_BASE_REF)`. It works the same way in make, but looks more like a normal shell variable expansion.

build.make
mauriciopoppe: I took the description of `PULL_BASE_REF` from https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md#custom-substitutions as the comment for this variable, I understand now why it might be dangerous

mauriciopoppe:
I've updated the PR to use `${PULL_BASE_REF}` instead of `$(PULL_BASE_REF)`

build.make
```diff
@@ -185,24 +185,24 @@ $(CMDS:%=push-multiarch-%): push-multiarch-%: check-pull-base-ref build-%
 		done; \
 		docker manifest push -p $(IMAGE_NAME):$$tag; \
 	}; \
-	if [ $(PULL_BASE_REF) = "master" ]; then \
+	if [ $${PULL_BASE_REF} = "master" ]; then \
```
pohly, line 188: You don't need `$$`, a single `$` is enough. Or is this intentional because you want to see `${PULL_BASE_REF}` in the make output when it dumps the command?

mauriciopoppe: I see, I thought I'd need to escape `$` with `$$` inside a Makefile target but I understand that variable expansion will happen before, I've updated the PR to use a single `$`

PR #173, opened 2021-09-03
Added WS2022 to 'CSI_PROW_BUILD_PLATFORMS'

nick5616:
This PR adds Windows Server 2022 to the build platforms. In the future, I will sync this with the relevant subtrees


mauriciopoppe:
The windows binary was created successfully but I was not able to check that this docker manifest includes the ltsc2022 version, to verify that this actually works I'd suggest making the following changes in https://github.com/kubernetes-csi/livenessprobe/pull/128

- remove this check temporary in your livenessprobe PR `./release-tools/verify-subtree.sh release-tools`, that might mean editing the `build.make` file in `release-tools`

pohly (approved):
/approve

PR #178, opened 2022-01-05
Increase build timeout

xing-yang:
Building 3 snapshot images are taking longer than an hour recently.  As a result, it timed out before finish building the snapshot validation webhook image.  That's why we only see snapshot-controller and CSI snapshotter sidecar images in the staging repo, but not the webhook one, when cutting release v5.0.0 for external-snapshotter.
https://k8s-testgrid.appspot.com/sig-storage-image-build#post-external-snapshotter-push-images

So we need to bump the timeout value.


msau42:
/lgtm
/approve
It would be nice too if we can get timestamps in the logs to see how long each step took.

xing-yang:
> It would be nice too if we can get timestamps in the logs to see how long each step took.

I'll open an issue to add timestamps.

PR #187, opened 2022-01-26
Remove Windows versions 1909 and 2004 which have reached EOL

mauriciopoppe:
/hold
/cc @jingxu97 @pohly @andyzhangx 

Removing support for some Windows versions that have reached EOL https://endoflife.date/windows, the versions that we still maintain are ltsc2019, 20H2 and ltsc2022.

I think we could merge this in 1.24.


andyzhangx (approved):
/lgtm

PR #190, opened 2022-02-14
support customized LDFLAGS in image build

andyzhangx:
would like to add more info in LDFLAGS in image build

### DIFF
--- a/build.make
+++ b/build.make
@@ -77,9 +77,9 @@ endif
 BUILD_PLATFORMS =
 
 # Add go ldflags using LDFLAGS at the time of compilation.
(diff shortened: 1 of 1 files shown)


pohly:
How is this meant to be used? During manually triggered image builds? When and why are those needed?

pohly:
You know that "make EXT_LDFLAGS=..." overrides the values in the Makefile?

pohly:
> if I define LDFLAGS in csi driver's Makefile, will it overiride? 

The last value is the one that gets used. So first build.make sets its value, then Makefile can override it. So what you have right now in your Makefile should work. Are you seeing something different?

PR #193, opened 2022-05-10
correct kube version condition check for default_csi_snapshotter_version func

humblec:
Additional note for reviewer:

The csi snapshotter would be picked as `v3.0.2` regardless of the KUBE version being latest due to the wrong condition check. This PR correct the same.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -351,7 +351,7 @@ configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_
 
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
 default_csi_snapshotter_version () {
-	if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" = "canary" ]; then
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -351,7 +351,7 @@ configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_
 
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
 default_csi_snapshotter_version () {
-	if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" = "canary" ]; then
+	if [ "${CSI_PROW_KUBERNETES_VERSION}" == "latest" ] || [ "${CSI_PROW_DRIVER_CANARY}" == "canary" ]; then
```
pohly, line 354: I'm confused. What does this have to do with "use latest version of the snapshot controller based on kube version"?

pohly:
The "why" part is missing - why is the old code incorrect and the new one correct? Where does it matter?

To me, the new version seems worse (bash syntax instead of POSIX), but perhaps I am missing something.

humblec:
> The "why" part is missing - why is the old code incorrect and the new one correct? Where does it matter?
> 
> To me, the new version seems worse (bash syntax instead of POSIX), but perhaps I am missing something.

Thanks @pohly , being POSIX conformant we can stick to that syntax. Closing this PR.

PR #195, opened 2022-06-09
fix testing with latest Kubernetes

pohly:
It failed for two reasons:
- CSIDriver v1beta1 as used in csi-driver-host-path v1.3.0 is no longer
  supported.
- GenericEphemeralVolume was removed in Kubernetes master.

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -196,7 +196,7 @@ kindest/node:v1.14.10@sha256:f8a66ef82822ab4f7569e91a5bccaf27bceee135c1457c512e5
 # If the deployment script is called with CSI_PROW_TEST_DRIVER=<file name> as
 # environment variable, then it must write a suitable test driver configuration
 # into that file in addition to installing the driver.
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -1304,7 +1304,11 @@ main () {
             delete_cluster_inside_prow_job non-alpha
         fi
 
-        if tests_need_alpha_cluster && [ "${CSI_PROW_E2E_ALPHA_GATES}" ]; then
+        # If the cluster for alpha tests doesn't need any feature gates, then we
+        # could reuse the same cluster as for the other tests. But that would make
+        # the flow in this script harder and wouldn't help in practice because
+        # we have separate Prow jobs for alpha and non-alpha tests.
+        if tests_need_alpha_cluster; then
```
jingxu97, line 1311: if we have different prow jobs for alpha tests, is it going to recreate the cluster here? just want to make sure for each job, we only need to create cluster once..

pohly: Each job will create and delete one cluster because either the one if path is taken or the other, but never both in the same job.

PR #197, opened 2022-08-03
support Kubernetes 1.25 + Ginkgo v2

pohly:
Kubernetes 1.25 is switching to Ginkgo v2 for the e2e.test suite.
This has two implications:
- we must built a matching ginkgo binary because communication
  between the ginkgo binary and library changed, affecting for
  example running of parallel tests: this is done by using the
  Kubernetes make rules to build the binary when using the Kubernetes
  e2e.test suite
- the ginkgo v2 JUnit file is slightly different: filter-junit.go
  now first tries the old format, then the new one

### DIFF
--- a/filter-junit.go
+++ b/filter-junit.go
@@ -35,10 +35,18 @@ var (
 )
 
 /*
- * TestSuite represents a JUnit file. Due to how encoding/xml works, we have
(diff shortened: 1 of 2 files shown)


pohly:
I verified locally that this works with Kubernetes master.

PR #199, opened 2022-08-05
prow.sh: bump Kubernetes to v1.22.0

pohly:
This update is necessary because the default driver
version (csi-driver-host-path v1.8.0) no longer supports Kubernetes v1.17.0 (it
needs the v1 CSIDriver API).

1.22 is the oldest supported version at this point. This matches what we are
using as base in Prow at the moment. v1.22.0 is what the current kind has
pre-built images for.


pohly:
nil pointer access in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-release-tools/199/pull-kubernetes-csi-release-tools-csi-test/1555477126818304000 during "make test"?! That seems unrelated.

PR #206, opened 2022-09-23
Changes to csi prow to run e2e tests in sidecars

RaunakShah:
Able to run on external-provisioner when tests are added to that repo:

```
Thu Sep 22 10:20:31 IST 2022 go1.18.3 /Users/raunakshah/go/src/src/github.com/kubernetes-csi/external-provisioner$ env KUBECONFIG=/Users/raunakshah/.kube/config KUBE_TEST_REPO_LIST= ginkgo -v -focus=Test1 /Users/raunakshah/go/src/pkg/csiprow.ProhizqqL5/e2e.test -- -report-dir /Users/raunakshah/go/src/pkg/csiprow.ProhizqqL5/artifacts
Sep 22 10:20:32.938: INFO: The --provider flag is not set. Continuing as if --provider=skeleton had been used.
Running Suite: E2E suite - /Users/raunakshah/go/src/pkg/csiprow.ProhizqqL5
==========================================================================
Random Seed: 1663822232

Will run 1 of 1 specs
------------------------------
Test1 test1
  test-1-A
  /Users/raunakshah/go/src/github.com/kubernetes-csi/external-provisioner/test/e2e/storage/provision.go:13

...

Ran 1 of 1 Specs in 0.034 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

```release-note
Add ability to run e2e tests written in sidecar repos
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -228,6 +228,9 @@ configvar CSI_PROW_E2E_VERSION "$(version_to_git "${CSI_PROW_KUBERNETES_VERSION}
 configvar CSI_PROW_E2E_REPO "https://github.com/kubernetes/kubernetes" "E2E repo"
 configvar CSI_PROW_E2E_IMPORT_PATH "k8s.io/kubernetes" "E2E package"
 
+# Local path for e2e tests. Set to "none" to disable.
+configvar CSI_PROW_SIDECAR_E2E_IMPORT_PATH "none" "CSI Sidecar E2E package"
+
 # csi-sanity testing from the csi-test repo can be run against the installed
 # CSI driver. For this to work, deploying the driver must expose the Unix domain
 # csi.sock as a TCP service for use by the csi-sanity command, which runs outside
@@ -282,13 +285,18 @@ tests_enabled () {
 sanity_enabled () {
     [ "${CSI_PROW_TESTS_SANITY}" = "sanity" ] && tests_enabled "sanity"
 }
+
+sidecar_tests_enabled () {
+  [ "${CSI_PROW_SIDECAR_E2E_IMPORT_PATH}" != "none" ]
+}
+
 tests_need_kind () {
     tests_enabled "parallel" "serial" "serial-alpha" "parallel-alpha" ||
-        sanity_enabled
+        sanity_enabled || sidecar_tests_enabled
 }
 tests_need_non_alpha_cluster () {
     tests_enabled "parallel" "serial" ||
-        sanity_enabled
+        sanity_enabled || sidecar_tests_enabled
 }
 tests_need_alpha_cluster () {
     tests_enabled "parallel-alpha" "serial-alpha"
@@ -351,6 +359,11 @@ configvar CSI_PROW_E2E_ALPHA_GATES "$(get_versioned_variable CSI_PROW_E2E_ALPHA_
 
 configvar CSI_PROW_E2E_GATES_LATEST '' "non alpha feature gates for latest Kubernetes"
 configvar CSI_PROW_E2E_GATES "$(get_versioned_variable CSI_PROW_E2E_GATES "${csi_prow_kubernetes_version_suffix}")" "non alpha E2E feature gates"
+
+# Focus for local tests run in the sidecar E2E repo. Only used if CSI_PROW_SIDECAR_E2E_IMPORT_PATH
+# is not set to "none". If empty, all tests in the sidecar repo will be run.
+configvar CSI_PROW_SIDECAR_E2E_FOCUS '' "tags for local E2E tests"
+configvar CSI_PROW_SIDECAR_E2E_SKIP '' "local tests that need to be skipped"
 
 # Which external-snapshotter tag to use for the snapshotter CRD and snapshot-controller deployment
 default_csi_snapshotter_version () {
@@ -942,6 +955,9 @@ install_e2e () {
         return
     fi
 
+    if sidecar_tests_enabled; then
+        run_with_go "${CSI_PROW_GO_VERSION_BUILD}" go test -c -o "${CSI_PROW_WORK}/e2e-local.test" "${CSI_PROW_SIDECAR_E2E_IMPORT_PATH}"
+    fi
     git_checkout "${CSI_PROW_E2E_REPO}" "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" "${CSI_PROW_E2E_VERSION}" --depth=1 &&
     if [ "${CSI_PROW_E2E_IMPORT_PATH}" = "k8s.io/kubernetes" ]; then
         patch_kubernetes "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" "${CSI_PROW_WORK}" &&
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -997,8 +1008,13 @@ run_e2e () (
     }
     trap move_junit EXIT
 
-    cd "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" &&
-    run_with_loggers env KUBECONFIG="$KUBECONFIG" KUBE_TEST_REPO_LIST="$(if [ -e "${CSI_PROW_WORK}/e2e-repo-list" ]; then echo "${CSI_PROW_WORK}/e2e-repo-list"; fi)" ginkgo -v "$@" "${CSI_PROW_WORK}/e2e.test" -- -report-dir "${ARTIFACTS}" -storage.testdriver="${CSI_PROW_WORK}/test-driver.yaml"
+    if [ "${name}" == "local" ]; then
+        cd "${GOPATH}/src/${CSI_PROW_SIDECAR_E2E_IMPORT_PATH}" &&
+        run_with_loggers env KUBECONFIG="$KUBECONFIG" KUBE_TEST_REPO_LIST="$(if [ -e "${CSI_PROW_WORK}/e2e-repo-list" ]; then echo "${CSI_PROW_WORK}/e2e-repo-list"; fi)" ginkgo -v "$@" "${CSI_PROW_WORK}/e2e.test" -- -report-dir "${ARTIFACTS}"
```
RaunakShah, line 1013: the `-storage.testdriver` parameter is not supported in sidecar repos

prow.sh
```diff
@@ -942,16 +949,20 @@ install_e2e () {
...
 
-    git_checkout "${CSI_PROW_E2E_REPO}" "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" "${CSI_PROW_E2E_VERSION}" --depth=1 &&
-    if [ "${CSI_PROW_E2E_IMPORT_PATH}" = "k8s.io/kubernetes" ]; then
-        patch_kubernetes "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" "${CSI_PROW_WORK}" &&
-        go_version="${CSI_PROW_GO_VERSION_E2E:-$(go_version_for_kubernetes "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" "${CSI_PROW_E2E_VERSION}")}" &&
-        run_with_go "$go_version" make WHAT=test/e2e/e2e.test "-C${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" &&
-        ln -s "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}/_output/bin/e2e.test" "${CSI_PROW_WORK}" &&
-        run_with_go "$go_version" make WHAT=vendor/github.com/onsi/ginkgo/ginkgo "-C${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" &&
-        ln -s "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}/_output/bin/ginkgo" "${CSI_PROW_BIN}"
+    if [ "${CSI_PROW_SIDECAR_E2E_IMPORT_PATH}" != "none" ]; then
```
pohly, line 952: This can be replaced with `sidecar_tests_enabled`, right?

pohly: I'm unclear about one thing regarding this PR: is the goal to run either a local E2E suite or the one from Kubernetes, or is the goal to support both in the same job?

Because of this if check here it seems to be the former, because `install_e2e` can only be used to install one suite, but not both. I think it would be better to support both suites in the same job.

RaunakShah: @pohly That makes sense. I've updated the code in the latest commit to run the tests in a single job. My original testing only ran the local tests, so I didn't realise that `install_e2e` is run only one (although now the `if` at the start of that function makes more sense :))

pohly (changes requested):

prow.sh
```diff
@@ -997,8 +1007,13 @@ run_e2e () (
...
 
-    cd "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" &&
-    run_with_loggers env KUBECONFIG="$KUBECONFIG" KUBE_TEST_REPO_LIST="$(if [ -e "${CSI_PROW_WORK}/e2e-repo-list" ]; then echo "${CSI_PROW_WORK}/e2e-repo-list"; fi)" ginkgo -v "$@" "${CSI_PROW_WORK}/e2e.test" -- -report-dir "${ARTIFACTS}" -storage.testdriver="${CSI_PROW_WORK}/test-driver.yaml"
+    if [ "${name}" == "local" ]; then
+        cd "${GOPATH}/src/${CSI_PROW_SIDECAR_E2E_IMPORT_PATH}" &&
+        run_with_loggers env KUBECONFIG="$KUBECONFIG" KUBE_TEST_REPO_LIST="$(if [ -e "${CSI_PROW_WORK}/e2e-repo-list" ]; then echo "${CSI_PROW_WORK}/e2e-repo-list"; fi)" ginkgo -v "$@" "${CSI_PROW_WORK}/e2e-local.test" -- -report-dir "${ARTIFACTS}"
+    else
+        cd "${GOPATH}/src/${CSI_PROW_E2E_IMPORT_PATH}" &&
+        run_with_loggers env KUBECONFIG="$KUBECONFIG" KUBE_TEST_REPO_LIST="$(if [ -e "${CSI_PROW_WORK}/e2e-repo-list" ]; then echo "${CSI_PROW_WORK}/e2e-repo-list"; fi)" ginkgo -v "$@" "${CSI_PROW_WORK}/e2e.test" -- -report-dir "${ARTIFACTS}" -storage.testdriver="${CSI_PROW_WORK}/test-driver.yaml"
+    fi
```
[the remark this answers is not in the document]
RaunakShah: Done. Do you think we can remove this part as well? 
```
KUBE_TEST_REPO_LIST="$(if [ -e "${CSI_PROW_WORK}/e2e-repo-list" ]; then echo "${CSI_PROW_WORK}/e2e-repo-list"; fi)
```

prow.sh
```diff
@@ -1310,6 +1325,14 @@ main () {
                         ret=1
                     fi
                 fi
+
+                if sidecar_tests_enabled; then
+                    if ! run_e2e local \
+                         -focus="some regex"; then
```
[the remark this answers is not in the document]
RaunakShah: Oops, this was from my testing. But actually I think we don't need any `focus` for local tests right? Ideally all tests from within the repo should be run.

prow.sh
RaunakShah: @pohly tests running in the sidecar repos would potentially want to enable some feature gates in the container. I think this file would need further changes to support that. If it runs in a single job, the feature gates would be enabled for the Kubernetes tests as well.

RaunakShah: Unless of course we undo the change after the local tests are complete

prow.sh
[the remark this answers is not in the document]
pohly: Usually yes, but there might be exceptions. As this script is going to be used by many different components, it's better to be flexible. Let's introduce a separate variable for it.

prow.sh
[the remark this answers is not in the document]
pohly: No, let's keep it. The local test might need the same image remapping as upstream Kubernetes. If it doesn't, then setting `KUBE_TEST_REPO_LIST` won't hurt.

prow.sh
RaunakShah: Does this apply to out of tree feature gates as well? We don't need to rebuild, just patch the particular sidecar manifest with the right arguments. 
Sidecar repos with alpha features would also contain e2e tests for those features. So certain feature gates in that sidecar only need to be enabled when running those tests, right?

prow.sh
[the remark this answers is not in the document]
RaunakShah: Done! Introduced `CSI_PROW_SIDECAR_E2E_FOCUS`

prow.sh
pohly: The feature gates for the driver are different, but isn't the driver also only getting deployed once for both E2E suites?

pohly: True. But as with cluster feature gates I prefer simplicity over flexibility here unless we know that it is needed.

RaunakShah: So will we need a change to this `if` to enable out-of-tree features in the sidecar manifest?

RaunakShah: Okay, I'll need to take a deeper look into where to update the feature flags. Potentially here - https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L695 but let me play around with it a bit

RaunakShah: Should we move ahead with this PR in the meanwhile? I don't think there's an obvious solution to adding features gates in a particular sidecar container.

xing-yang:
@RaunakShah Can you add a release note?

RaunakShah:
>  Can you add a release note?
Done!

prow.sh
pohly: We can do that. Can you squash commits and then try out that commit in some sidecar repo where a local E2E suite is defined? In other words, create a test PR in some other repo?

It will fail the unit check for locally modified `release-tools`, but that's fine.

prow.sh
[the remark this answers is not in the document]
jsafrane: Please introduce also CSI_PROW_SIDECAR_E2E_SKIP, we often need to skip test for alpha features or disruptive tests.

prow.sh
RaunakShah: Created a PR here - https://github.com/kubernetes-csi/external-provisioner/pull/798
The local test is just a skeleton to prove that the tests within that repo will run with these changes.

prow.sh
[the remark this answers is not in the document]
RaunakShah: Done!

prow.sh
```diff
@@ -264,7 +267,7 @@ configvar CSI_PROW_DEP_VERSION v0.5.1 "golang dep version to be used for vendor
 # thus only makes sense in repos which provide their own CSI
 # driver. Repos can enable sanity testing by setting
 # CSI_PROW_TESTS_SANITY=sanity.
-configvar CSI_PROW_TESTS "unit parallel serial $(if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ]; then echo parallel-alpha serial-alpha; fi) sanity" "tests to run"
+configvar CSI_PROW_TESTS "local unit parallel serial $(if [ "${CSI_PROW_KUBERNETES_VERSION}" = "latest" ]; then echo parallel-alpha serial-alpha; fi) sanity" "tests to run"
```
pohly, line 270: Does adding "local" here have any effect? I don't think so. Can you remove it?

pohly (changes requested):
I did another pass and only found one issue ("local" in `CSI_PROW_TESTS`).

prow.sh
RaunakShah: Removed!

pohly (approved):
/lgtm

PR #208, opened 2022-11-10
Skip SELinux tests in CI by default

jsafrane:
Currently, no CI job has SELinux configured, so skip all tests that require it by default.. When a job supports SELinux, it can set `CSI_PROW_E2E_SKIP='Disruptive'` in its prow yaml.

Fixes: https://github.com/kubernetes-csi/csi-driver-host-path/issues/376
Needs https://github.com/kubernetes/kubernetes/pull/113809 merged first.

```release-note
NONE
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -381,7 +381,7 @@ configvar CSI_SNAPSHOTTER_VERSION "$(default_csi_snapshotter_version)" "external
 # whether they can run with the current cluster provider, but until
 # they are, we filter them out by name. Like the other test selection
 # variables, this is again a space separated list of regular expressions.
-configvar CSI_PROW_E2E_SKIP 'Disruptive' "tests that need to be skipped"
+configvar CSI_PROW_E2E_SKIP '\[Disruptive\]|\[Feature:SELinux\]' "tests that need to be skipped"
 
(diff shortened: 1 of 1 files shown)


prow.sh
```diff
@@ -381,7 +381,7 @@ configvar CSI_SNAPSHOTTER_VERSION "$(default_csi_snapshotter_version)" "external
 # whether they can run with the current cluster provider, but until
 # they are, we filter them out by name. Like the other test selection
 # variables, this is again a space separated list of regular expressions.
-configvar CSI_PROW_E2E_SKIP 'Disruptive' "tests that need to be skipped"
+configvar CSI_PROW_E2E_SKIP 'Disruptive|\[Feature:SELinux\]' "tests that need to be skipped"
```
bertinatto, line 384: I think the correct form would be `\[Disruptive\]|\[Feature:SELinux\]`.
Example: https://github.com/kubernetes/test-infra/blob/62bfb1841cef4df5170795067116a94a0a891f01/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml#L44

jsafrane: Fixed

PR #215, opened 2022-12-23
Set CSI hostpath driver version to master if canary run

RaunakShah:
Even if `CSI_PROW_HOSTPATH_CANARY` is set to `canary`, the prow job runs with CSI hostpath driver version set to the default (currently `v1.8.0`)
This PR sets it to master for canary runs. 

Example prow run with the issue - https://prow.k8s.io/log?container=test&id=1606218480988721152&job=pull-kubernetes-csi-external-provisioner-canary

```release-note
Set CSI hostpath driver version to master for canary prow jobs
```


pohly:
Why can't a canary job that sets `CSI_PROW_DRIVER_CANARY` not also set `CSI_PROW_DRIVER_VERSION=master`?

This change doesn't seem necessary.

PR #224, opened 2023-04-25
Update cloudbuild image

msau42:
Try to fix build failures:
https://storage.googleapis.com/kubernetes-jenkins/logs/post-external-provisioner-push-images/1650605297175105536/build-log.txt

Looks like go 1.20 added a new requirement to dynamically link to libresolv.so, and our image builder is based on alpine which doesn't support it:
https://github.com/golang/go/issues/59305


pohly (approved):
/lgtm
/approve

PR #227, opened 2023-05-16
Check CSI sidecar supported versions

coulof:
That script is a helper to update the tables from : https://kubernetes-csi.github.io/docs/sidecar-containers.html

The next goal is to generate the tables from the release pages but... we have to agree upon the metadata and fix the typos :-) (cf. https://github.com/kubernetes-csi/external-attacher/releases with `docker pull docker pull registry.k8s.io/sig-storage/csi-attacher:v4.3.0`)

### DIFF
--- a/contrib/get_supported_version_csi-sidecar.py
+++ b/contrib/get_supported_version_csi-sidecar.py
@@ -1,0 +1,134 @@
+import argparse
+import datetime
+import re
+from collections import defaultdict
+import subprocess
+import shutil
+from dateutil.relativedelta import relativedelta
+
+# Check that the `gh` command is in the path
+def check_gh_command():
+    if not shutil.which('gh'):
+        print("Error: The `gh` command is not available in the PATH.")
+        print("Please install the GitHub CLI (https://cli.github.com/) and try again.")
+        exit(1)
+
+# humanize duration outputs
+def duration_ago(dt):
+    delta = relativedelta(datetime.datetime.now(), dt)
+    if delta.years > 0:
+        return f"{delta.years} year{'s' if delta.years > 1 else ''} ago"
+    elif delta.months > 0:
+        return f"{delta.months} month{'s' if delta.months > 1 else ''} ago"
+    elif delta.days > 0:
+        return f"{delta.days} day{'s' if delta.days > 1 else ''} ago"
+    elif delta.hours > 0:
+        return f"{delta.hours} hour{'s' if delta.hours > 1 else ''} ago"
+    elif delta.minutes > 0:
+        return f"{delta.minutes} minute{'s' if delta.minutes > 1 else ''} ago"
+    else:
+        return "just now"
+
+def parse_version(version):
+    # version pattern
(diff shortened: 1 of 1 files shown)


ggriffiths (approved):
/lgtm, looks good to me! Thanks for adding this!

coulof:
Thanks @ggriffiths.
I think this needs a /approve too ;-)

contrib/get_supported_version_csi-sidecar.py
```diff
@@ -0,0 +1,170 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import datetime
+import re
+from collections import defaultdict
+import subprocess
+import shutil
+from dateutil.relativedelta import relativedelta
```
mauriciopoppe, line 21: If this is an external dependency I think it needs to have a `pip install` command somewhere but it's a dev only tool so I think it's good.

Also consider that this file is going to be synced to all the CSI repos.

mauriciopoppe (commented):
Thanks for taking the time to write this tool. If there's tooling outside the usual bash scripts/makefile scripts I think all of our repos are written in go so it'd make sense to write tooling in go too.

ggriffiths:
> Thanks @ggriffiths. I think this needs a /approve too ;-)

Yes, you'll need @jsafrane for an approval review. I can only provide LGTM :)

ggriffiths:
> Thanks for taking the time to write this tool. If there's tooling outside the usual bash scripts/makefile scripts I think all of our repos are written in go so it'd make sense to write tooling in go too.

I think there are some K8s scripts out there written in a Python. It's a much better language than bash or go for this use case.

contrib/get_supported_version_csi-sidecar.py
coulof: `collections` is a built-in ; `dateutil` is an external dependency.

Next time I work on it I will replace it with just built-in ;-)

jsafrane:
/approve
sorry it took so long

contrib/get_supported_version_csi-sidecar.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2023 The Kubernetes Authors.
```
mauriciopoppe, line 1: Some repos have codespell enabled as a github action which failed when this change was synced to the other repos, could you please review these errors?

```
Resulting CLI options  --check-filenames --skip ./.git,./.github/workflows/codespell.yml,.git,*.png,*.jpg,*.svg,*.sum,./vendor,go.sum,./release-tools/prow.sh
3
Error: ./release-tools/contrib/get_supported_version_csi-sidecar.py:35: ouputs ==> outputs
Error: ./release-tools/contrib/get_supported_version_csi-sidecar.py:101: relase ==> release
```

mauriciopoppe: addressed in https://github.com/kubernetes-csi/csi-release-tools/pull/229

PR #243, opened 2023-11-22
pull-test.sh: avoid "git subtree pull" error

pohly:
Without a prior `git reset --hard`, `git subtree pull` fails with:

    fatal: working tree has modifications.  Cannot add.

Debugging with -d doesn't show any explanation and neither `git diff` nor `git diff-index` show any modifications. The `git reset --hard` shouldn't be necessary.


jsafrane:
/lgtm
thanks a lot for debugging this!

xing-yang:
Thanks @pohly !

PR #247, opened 2023-12-22
Scripts for patch releases

msau42:
Added a script to update specific go modules, and fixed the release notes script to not depend on discover mode which seems to be broken.

### DIFF
--- a/generate-patch-release-notes.sh
+++ b/generate-patch-release-notes.sh
@@ -48,7 +48,7 @@ function gen_patch_relnotes() {
   rm out.md || true
   rm -rf /tmp/k8s-repo || true
   GITHUB_TOKEN="$CSI_RELEASE_TOKEN" \
-  release-notes --discover=patch-to-latest --branch="$2" \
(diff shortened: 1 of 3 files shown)


generate-patch-release-notes.sh
```diff
@@ -48,7 +48,7 @@ function gen_patch_relnotes() {
   rm out.md || true
```
carlory, line 48: after I run this script, it doesn't cleanup the out.md.

```
(⎈|kind-kind:N/A)➜  csi-driver-host-path git:(changelog-release-1.12) ✗ git status
On branch changelog-release-1.12
Your branch is ahead of 'upstream/release-1.12' by 1 commit.
  (use "git push" to publish your local commits)

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	CHANGELOG/out.md

nothing added to commit but untracked files present (use "git add" to track)
```

msau42: fixed

carlory:
/remove-lifecycle stale
/lgtm
/hold 

feel free to cancel the hold

Issue #248, opened 2024-02-16
CSI CI starts to fail after adding K8s 1.29 jobs


xing-yang:
After adding jobs for K8s v1.29, CI started to fail.

https://testgrid.k8s.io/sig-storage-csi-ci#1.29-test-on-1.29

https://github.com/kubernetes-csi/external-snapshotter/pull/1012#issuecomment-1948678219

xing-yang:
For 1.28 and earlier jobs, it uses CSI_PROW_GO_VERSION_BUILD "1.21.5" defined in prow.sh. We fixed a similar issue earlier with this PR: https://github.com/kubernetes-csi/csi-release-tools/pull/246/files.

For the latest K8s version in CI, it tries to use the `minimum_go_version` in hack/lib/golang.sh. For K8s 1.29, the `minimum_go_version` is 1.21.

https://github.com/kubernetes-csi/csi-release-tools/blob/master/prow.sh#L562
```
go_version_for_kubernetes () (
    local path="$1"
    local version="$2"
    local go_version

    # We use the minimal Go version specified for each K8S release (= minimum_go_version in hack/lib/golang.sh).
    # More recent versions might also work, but we don't want to count on that.
    go_version="$(grep minimum_go_version= "$path/hack/lib/golang.sh" | sed -e 's/.*=go//')"
    if ! [ "$go_version" ]; then
        die "Unable to determine Go version for Kubernetes $version from hack/lib/golang.sh."
    fi
    # Strip the trailing .0. Kubernetes includes it, Go itself doesn't.
    # Ignore: See if you can use ${variable//search/replace} instead.
    # shellcheck disable=SC2001
    go_version="$(echo "$go_version" | sed -e 's/\.0$//')"
    echo "$go_version"
)
```
However, go1.21.linux-amd64.tar.gz does not exist.
```
Fri Feb 16 15:00:38 UTC 2024 go1.22.0 /home/prow/go/pkg/csiprow.XBzTGe3OIr/src/kubernetes$ curl --fail --location https://dl.google.com/go/go1.21.linux-amd64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0  1449    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

[...]

PR #249, opened 2024-02-21
Use .go-version to get Kubernetes go version

jsafrane:
Kubernetes publishes `.go-version` since v1.26.1. Use it to get Kubernete's go version instead of parsing a shell script. Fall back to parsing of hack/lib/golang.sh if `.go-version` does not exist.

Fixes: https://github.com/kubernetes-csi/csi-release-tools/issues/248

```release-note
Use .go-version file from github.com/kubernetes/kubernetes (k/k) to detect go version that k/k needs.
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -564,7 +564,15 @@ go_version_for_kubernetes () (
     local version="$2"
     local go_version
 
-    # We use the minimal Go version specified for each K8S release (= minimum_go_version in hack/lib/golang.sh).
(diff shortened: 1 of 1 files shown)


jsafrane:
I updated the PR to fall back to the original behavior when `.go-version` does not exist.

xing-yang:
1.29 job passed: https://github.com/kubernetes-csi/external-snapshotter/pull/1018

Issue #235, opened 2023-09-11
Golang version update process


msau42:
Considering adopting a form of the [kubernetes golang backport process](https://kubernetes.io/blog/2023/04/06/keeping-kubernetes-secure-with-updated-go-versions/) to CSI releases as well.

A summary of the process is:
1. In the next minor releases, update to the latest go version that has been out for at least 3 months
2. After a month of soak time on the new minor releases, then backport the go version to previous release branches

Before we can realistically adopt this, we need to get our release process into a more automated state so that we can regularly release patch versions. @sunnylovestiramisu is investigating that.

PR #250, opened 2024-04-23
Update go to 1.22.2

jsafrane:
Update go version to be in sync with Kubernetes 1.30, which uses 1.22.2.

cc @xing-yang @msau42 

```release-note
Go updated to 1.22.2
```


jsafrane:
Go 1.22.0 was released on 2024-02-06, which is less than 3 months old, i.e. less than required by https://github.com/kubernetes-csi/csi-release-tools/issues/235. Still, I think it's useful to be in sync with Kubernetes 1.30.

PR #251, opened 2024-05-10
Add test-logcheck target

jsafrane:
The new `test-logcheck` + `verify-logcheck.sh` script checks for correct klog usage.

It's optional and must be explicitly requested in the main Makefile by adding:
```
test: test-logcheck
```

```release-note
Added support for structured and contextual logging check
```


bells17:
Thank you!
/lgtm

PR #257, opened 2024-06-26
chore: add CSI_PROW_SIDECAR_E2E_PATH

jakobmoellerdev:
adds CSI_PROW_SIDECAR_E2E_PATH, an alternative to CSI_PROW_SIDECAR_E2E_IMPORT_PATH that can be used if the go module referenced in the import path has a /v{x} suffix, making it differing from the directory of the module.

This allows https://github.com/kubernetes-csi/external-provisioner/pull/1236#issuecomment-2185900044 to be fixed by referencing it explicitly


jakobmoellerdev:
@msau42 would be cool if you could double check this?

jakobmoellerdev:
@msau42 sorry for the necro, ping :)

msau42:
Hm I guess we should be updating to golang 1.22.5 as well.

jakobmoellerdev:
ack, you want me to bump in this PR?

PR #258, opened 2024-08-19
Always enable race detection while running tests

gnufied:
Always enable race detection while running tests. This is useful for detecting simpler data races in code.

```release-note
Always enable race detection while running tests
```

### DIFF
--- a/build.make
+++ b/build.make
@@ -62,9 +62,9 @@ IMAGE_NAME=$(REGISTRY_NAME)/$*
 
 ifdef V
 # Adding "-alsologtostderr" assumes that all test binaries contain glog. This is not guaranteed.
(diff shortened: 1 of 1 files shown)


build.make
```diff
@@ -62,9 +62,9 @@ IMAGE_NAME=$(REGISTRY_NAME)/$*
 
 ifdef V
 # Adding "-alsologtostderr" assumes that all test binaries contain glog. This is not guaranteed.
-TESTARGS = -v -args -alsologtostderr -v 5
+TESTARGS = -race -v -args -alsologtostderr -v 5
```
xing-yang, line 65: Can you add a comment for the race flag?

PR #261, opened 2024-10-29
Bump golang to 1.23.1

huww98:
external-snapshotter is already using 1.23.1 in go.mod

```release-note
Bump Golang to 1.23.1
```

### DIFF
--- a/prow.sh
+++ b/prow.sh
@@ -86,7 +86,7 @@ configvar CSI_PROW_BUILD_PLATFORMS "linux amd64 amd64; linux ppc64le ppc64le -pp
 # which is disabled with GOFLAGS=-mod=vendor).
 configvar GOFLAGS_VENDOR "$( [ -d vendor ] && echo '-mod=vendor' )" "Go flags for using the vendor directory"
 
-configvar CSI_PROW_GO_VERSION_BUILD "1.22.5" "Go version for building the component" # depends on component's source code
+configvar CSI_PROW_GO_VERSION_BUILD "1.23.1" "Go version for building the component" # depends on component's source code
 configvar CSI_PROW_GO_VERSION_E2E "" "override Go version for building the Kubernetes E2E test suite" # normally doesn't need to be set, see install_e2e
 configvar CSI_PROW_GO_VERSION_SANITY "${CSI_PROW_GO_VERSION_BUILD}" "Go version for building the csi-sanity test suite" # depends on CSI_PROW_SANITY settings below
 configvar CSI_PROW_GO_VERSION_KIND "${CSI_PROW_GO_VERSION_BUILD}" "Go version for building 'kind'" # depends on CSI_PROW_KIND_VERSION below


huww98:
BTW, do you think we should `export GOTOOLCHAIN=local` in run_with_go? If not, this function is meaningless because go will download the version specified in the `go.mod` itself.

huww98:
/cc @xing-yang @jsafrane
PTAL

jsafrane:
> BTW, do you think we should export GOTOOLCHAIN=local in run_with_go? If not, this function is meaningless because go will download the version specified in the go.mod itself. Or should we just remove run_with_go function and rely on go command to manage the version?

What approach does the main kubernetes/kubernetes use? Is there anything we could copy?

pohly:
Manually downloading the right toolchain archive pre-dates the download mechanism in the Go toolchain itself. Nowadays using the toolchain to download the desired version is the right way to go.

huww98:
Can we get this PR merged? Or should we fix the (seems unrelated) CI failure first?

jsafrane:
We need the CI to be green before merge to make sure this PR does not break anything.

jsafrane:
I am staring at the CI logs and they don't make much sense to me.

pohly:
I suspect it fails because one or both repos are not completely present locally:
> --filter=blob:none will filter out all blobs (file contents) until needed by Git

This is used in https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-csi_csi-release-tools/261/pull-kubernetes-csi-release-tools-external-snapshotter/1854867176855441408/clone-log.txt.

I'm not sure yet how to change this - will ask on Slack.

huww98:
> When cloning or fetching from a partial repository (i.e., one itself cloned with --filter), the server-side upload-pack may need to fetch extra objects from its upstream in order to complete the request. By default, upload-pack will refuse to perform such a lazy fetch

https://git-scm.com/docs/git-upload-pack

So I think set GIT_NO_LAZY_FETCH=0 should fix this.

pull-test.sh
```diff
@@ -26,7 +26,7 @@ CSI_RELEASE_TOOLS_DIR="$(pwd)"
 # Update the other repo.
 cd "$PULL_TEST_REPO_DIR"
 git reset --hard # Shouldn't be necessary, but somehow is to avoid "fatal: working tree has modifications.  Cannot add." (https://stackoverflow.com/questions/3623351/git-subtree-pull-says-that-the-working-tree-has-modifications-but-git-status-sa)
-git subtree pull --squash --prefix=release-tools "$CSI_RELEASE_TOOLS_DIR" master
+GIT_NO_LAZY_FETCH=0 git subtree pull --squash --prefix=release-tools "$CSI_RELEASE_TOOLS_DIR" master
```
jsafrane, line 29: Please add a comment why we need GIT_NO_LAZY_FETCH and perhaps put it to  a standalone line

jsafrane:
Almost, please also revendor.

pull-test.sh
huww98: Rebased onto #264

huww98:
@pohly @jsafrane  Can you approve this PR? So that we can merge these two commits at once.

PR #266, opened 2024-12-03
Bump csi-sanity to 5.3.1

jsafrane:
5.3.1 is the latest version.
```release-note
NONE
```


xing-yang:
Not sure why we are getting the CI failure.  It happens with every PR here.

Issue #304, opened 2026-05-12
verify-subtree.sh does not work with tide/merge-method-squash


jsafrane:
See PR https://github.com/kubernetes-csi/csi-lib-utils/pull/213. It has a regular commit that updates Kubernetes libraries and two commits with csi-release-tools subtree update. The test verify-subtree.sh has passed there just fine, because of these separate commits.

But because of `/label tide/merge-method-squash`, these commits got squashed into one and verify-subtree.sh does not pass on the result. See https://github.com/kubernetes-csi/csi-lib-utils/pull/214 / its failed [pull-kubernetes-csi-csi-lib-utils](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-csi_csi-lib-utils/214/pull-kubernetes-csi-csi-lib-utils/2054128607936122880).

We need to fix the test to tolerate the squashed PR merges - people will use use it, unless we forcibly turn it off.

jsafrane:
Ideas:

* Improve the test to make sure **the whole PR** contains only release-tools changes to prevent future `tide/merge-method-squash` with other random commits in the same PR.
* Or check that just the commits in the PR are valid and tolerate previously squashed commits.
* Or check that changes in `release-tools/` are valid and ignore other changes. I.e. check that "`Squashed 'release-tools/' changes from 74502e54..e019f2a7` in the commit message  corresponds to the current `release-tools/` content somehow, but ignore any content outside of this directory.
* Disable `tide/merge-method-squash` in CSI repos.
* Or anything else.

pohly:
> Disable tide/merge-method-squash in CSI repos.

I looked and could only find a configuration option for Tide which flips the default: https://docs.prow.k8s.io/docs/components/core/tide/config

I'm still not a fan of `/label tide/merge-method-squash`. I've had cases where reviewers added it to my PRs where I had intentionally kept changes in separate commits, thus changing what I wanted to get merged. That kind of power should not be given to all Kubernetes members.

It is useful in repos with frequent contributions from contributors who don't know how to do manual squashing. In repos like the ones using csi-lib-utils I'd expect contributors to be able to do squashing manually and provide clean commit messages, so an option to not allow `merge-method-squash` in those repos IMHO would be useful.

pohly:
The GitHub settings for a repository have an option to disable squash and rebase merging. Tide then may fail to use it, but that's okay (remove the label and it should proceed).

PR #305, opened 2026-05-13
Verify subtree contents instead of merge history

torredil:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

[verify-subtree.sh](https://github.com/kubernetes-csi/csi-release-tools/blob/master/verify-subtree.sh) rejects any non-merge commits that touches `release-tools/` on the assumption that subtree pulls always land as merge commits. This assumption breaks when, for example, a PR carrying a subtree pull is squash-merged. In that scenario, the result is a single non-merge commit, which causes the test to fail and every subsequent PR is blocked even though `release-tools/` matches upstream.

This PR fixes the script to check the property we actualy care about here, which is `release-tools/` having the same contents as upstream. We do this by fetching the SHA and comparing the trees - if they match, good, if not, print a diff so the reviewer/contributor can see what's off. As a bonus this also catches edits snuck in via merge commits which the old check was letting through (see notes below).

**Which issue(s) this PR fixes**:

This happed on https://github.com/kubernetes-csi/csi-lib-utils/pull/213 and is currently blocking https://github.com/kubernetes-csi/csi-lib-utils/pull/214.

Fixes # https://github.com/kubernetes-csi/csi-release-tools/issues/304

**Special notes for your reviewer**:

For testing, i ran the new script against consuming repo's master and most recent release branch and it surfaced two issues:

  1. [`kubernetes-csi/csi-proxy/release-tools/prow.sh#L1267`](https://github.com/kubernetes-csi/csi-proxy/blob/master/release-tools/prow.sh#L1267) - diff at `"BUILD_PLATFORMS=${CSI_PROW_BUILD_PLATFORMS}"`
  2. [`kubernetes-csi/csi-driver-smb/release-tools/build.make#L156`](https://github.com/kubernetes-csi/csi-driver-smb/blob/master/release-tools/build.make#L156) and [`#L195-L204`](https://github.com/kubernetes-csi/csi-driver-smb/blob/master/release-tools/build.make#L195-L204). - Windows HostProcess container build steps diff.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

### DIFF
--- a/verify-subtree.sh
+++ b/verify-subtree.sh
@@ -14,28 +14,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script verifies that the content of a directory managed
-# by "git subtree" has not been modified locally. It does that
-# by looking for commits that modify the files with the
-# subtree prefix (aka directory) while ignoring merge
-# commits. Merge commits are where "git subtree" pulls the
-# upstream files into the directory.
+# Verify that DIR is a clean copy of upstream csi-release-tools.
 #
-# Theoretically a developer can subvert this check by modifying files
-# in a merge commit, but in practice that shouldn't happen.
+# We read the upstream SHA from the "git-subtree-split" line that
+# "git subtree" writes into the pull commit's message, fetch that
+# SHA, and require its tree to equal HEAD:DIR.
 
 DIR="$1"
-if [ ! "$DIR" ]; then
+if [ -z "$DIR" ]; then
     echo "usage: $0 <directory>" >&2
     exit 1
 fi
 
-REV=$(git log -n1 --remove-empty --format=format:%H --no-merges -- "$DIR")
-if [ "$REV" ]; then
-    echo "Directory '$DIR' contains non-upstream changes:"
-    echo
-    git log --no-merges -- "$DIR"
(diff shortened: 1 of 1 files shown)


torredil:
PTAL: @jsafrane @bswartz @pohly - thanks!

torredil:
cc: @xing-yang

pohly:
I think ensuring that the history of csi-release-tools changes is preserved in consumers is worthwhile. Relaxing the check to just "the end result must be the same" drops that.

Instead, https://github.com/kubernetes-csi/csi-release-tools/issues/304 can be avoided by disabling the squash and rebase methods in the CSI repos.

torredil:
@pohly We can have both 🙂  

This check verifies correctness directly instead of inferring it from commit shape, which is ideal, and disabling squash at the repo level enforces the merge strategy. What do you think?  either way, I'm happy to disagree and commit if closing this and relying on consumer repos to keep squash disabled is your best recommendation.

jsafrane:
I think we can have both too. I am adding an item to CSI call to discuss if we're disabling squash-merge in our repos.

/lgtm
/approve

PR #319, opened 2026-09-21
Use gofmt from GOTOOLCHAIN

jsafrane:
When using `gofmt`, use the one from GOTOOLCHAIN / CSI_PROW_GO_VERSION_BUILD instead of the one in the system or container image. The system one can be too new or too old.

See https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-csi_external-provisioner/1584/pull-kubernetes-csi-external-provisioner-unit/2101952853458292736 - CI correctly used go 1.26.6 from `CSI_PROW_GO_VERSION_BUILD=1.26.6` to build & unit-test the PR, but it used gofmt 1.27.1 for `make test-fmt`.

```release-note
NONE
```


jsafrane:
cc @gnufied @xing-yang PTAL

I have a test PR to confirm it works: https://github.com/kubernetes-csi/external-provisioner/pull/1586
`test-fmt` succeeds, only `test-subtree` fails because this PR is not yet merged.

